PR #4, opened 2018-11-04
Bugfixes

Bernator:
Bugfix, find asap in association Table not working correctly
Bugfix, WriteRequest of last Object in Object Table was not processed


thelsing:
The fix in bau_systemB.cpp is fine. But in the association table the highbyte is the tsap and lowbyte the asap (see 3/5/1 p. 157)
ntohs must be removed because the []-operator does it already. Can you change this please?

thelsing:
I just changed the files manually.

Bernator:
Thx

PR #7, opened 2018-11-21
Update knx-bme680.ino

henfri:
Made Pins and i2cAdress configurable


thelsing:
Thanks for the pull request. But I don't want to make the examples more complicated. But I should probably add some api documentation.

Issue #6, opened 2018-11-08
Separate directories for different platforms


nanosonde:
Hi!
I think it would be better to move the platform specific stuff to separate folders.

What do you think?

However, I am happy that you have removed the submodule stuff. :+1:

thelsing:
There are only 2 platform specific files per platform. (Well 3 for linux). So I don't see the need to split them up. But if you submit a pull request I will merge it ;)

nanosonde:
Ok, currently there are only those 3 platforms, but who knows. ;)

What I have in mind is something like this:
https://github.com/bluekitchen/btstack

thelsing:
Looks fine.

nanosonde:
I have closed the PR as I do not want to put any more effort into this for now as I am not sure if you really like the idea.

Just take it as an example of how I wanted to do it. However, I am not sure if it fits your VisualGDB development approach.

thelsing:
You are a bit too impatient. You created the pull request 4 days ago and closed it 2 days ago. I do this for fun and have a life too. So it can take some time till I check a pull request in detail.

nanosonde:
Sorry for the hassle. It is not related to the time frame in anyway, but more related to my activities and priorities of those. 
I have changed them spontaneously.

So again: no offense!

PR #11, opened 2019-01-03
platform and port specific files into seperate directories

nanosonde:
This pull request puts platform and port specific files into seperate directories as suggested here:
https://github.com/thelsing/knx/issues/6

It is basically the structure that btstack uses.
Arduino is just another platform next to the linux platform.
Maybe one should also add a similar Makefile like [this](https://github.com/bluekitchen/btstack/blob/master/port/arduino/Makefile) one which adds a convient way to generate the knx arduino library.

knxPython (from my point of view) is just another port of the knx stack.

Visualstudio files have also been moved and the paths were updated.


nanosonde:
Looking through the esp_platform.* and samd_platform.* files I realized that it does not make sense to put them into additional directories as they all use stuff from the arduino framework.

Another port and platform I am currently looking at is ESP32 (without Wifi, only BLE) with ESP-IDF and RTOS.

nanosonde:
Are you going to consider this PR?
Otherwise, I will stop working on it.

thelsing:
I have not testet it yet. If my usual workflow still works I'll merge this. The repository can't be used as a arduino lib directly after this, right?

nanosonde:
That's correct. It won't work as arduino lib anymore directly.
Therefore I have provided in the port/arduino folder. Just run it in the folder and it creates a ZIP file which contains all arduino source files in a single directory.

My suggestion would be to setup Travis-CI for this repository and whenever you create a tagged release then CI runs this make file and uploads the ZIP file to the Github release page.

PR #13, opened 2019-05-15
bugfix virtual void readProperty(PropertyID id, uint32_t start, uint32_t count, uint8_t *data)

Maggyver:


thelsing:
merged thanks

PR #14, opened 2019-05-17
features added

Maggyver:


thelsing:
Merged, Thanks. I change the Serial.print to print. Those functions are implemented in the *_platform.cpp file and write to the correct serial port for the board.

PR #15, opened 2019-05-19
circuit of the led adapted for different boards

Maggyver:


thelsing:
This will break the led for all non ESP8266 platforms. Just add a member variable and getters/setters to KnxFacade. So you can set something like knx.ledActiveOn(LOW);

thelsing:
This is still not exactly what I had in mind but it's close enough. I'll merge this thanks.
Some coding style things for next time:
- spaces instead of tabs for indentation.
- no braces for single line if else
- make sure you that you don't undo my changes in the merge

PR #16, opened 2019-05-20
Update knx-demo.ino

Maggyver:


Maggyver:
Mistake, please delete

PR #19, opened 2019-07-13
Update dptconvert.cpp

Maggyver:
Not verified.

### DIFF
--- a/src/knx/dptconvert.cpp
+++ b/src/knx/dptconvert.cpp
@@ -1809,7 +1809,7 @@ void float16ToPayload(uint8_t* payload, int payload_length, int index, double va
     int exponent = 0;
     for (; v < -2048.0f; v /= 2) exponent++;
     for (; v > 2047.0f; v /= 2) exponent++;
-    long m = round(v) & 0x7FF;
+    short m = roundf(v) & 0x7FF;
     short msb = (short) (exponent << 3 | m >> 8);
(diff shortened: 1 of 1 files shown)


src/knx/dptconvert.cpp
```diff
@@ -1713,9 +1713,11 @@ double float16FromPayload(const uint8_t* payload, int index)
 {
     uint16_t mantissa = unsigned16FromPayload(payload, index) & 0x87FF;
     if (mantissa & 0x8000)
-        return ((~mantissa & 0x07FF) + 1.0) * -0.01 * (1 << ((payload[index] >> 3) & 0x0F));
-
-    return mantissa * 0.01 * (1 << ((payload[index] >> 3) & 0x0F));
+        //return ((~mantissa & 0x07FF) + 1.0) * -0.01 * (1 << ((payload[index] >> 3) & 0x0F));
+        return ((~mantissa & 0x07FF) + 1.0) * -0.01 * pow(2.0, ((payload[index] >> 3) & 0x0F));
```
thelsing: Is this necessary? Shifting should produce the same result. And it should be much faster

Maggyver: Was just for testing, I'll delete that right away. But the way value to data point is fine or are there still problems?

thelsing:
Obsolete.

PR #20, opened 2019-07-14
async tpuart_data_link_layer

Bernator:
- split frame tx and confirm rx in an async manner
- add frame byte timeout
- add tx confirm timeout
- add crc check


thelsing:
Thank you for the pull request. Could you resolve the conflict please?
Other points:
- Please fix the whitespaces. Spaces should be used instead of tabs. At the moment you use both.
- The change of Serial1 to Serial2 in samd_platform probably depends on the board you use. Instead of changing the default could you add a method to change the Serial to the samd-Platform class and expose the platform in knx_facade.h? (Preferably in an other pull request.)

I still have o look at the details later.

Bernator:
sorry da ging zuviel mit, hab einen neuen pull request gemacht: 
https://github.com/thelsing/knx/pull/21

PR #22, opened 2019-07-23
bugfix, print not allowed in constructor

Bernator:
if print is used in any contructor __libc_init_array() will fail


thelsing:
The problem is that Platform is contructed before setup() so Serial is not set up yet. Maybe print in samd_platform.cpp should check this. But I'll merge this workaround.

PR #24, opened 2019-07-23
add method to change uart for tpuart communication (default Serial1)

Bernator:


thelsing:
I think it would be better to create a common baseclass ArduinoPlatform for SamdPlatform and EspPlatform and put the selectSerial and all the other common arduino stuff there.

Bernator:
obsolete

PR #28, opened 2019-08-17
 DPT16 (to bus) implemented

mumpf:
Hi, I added also some code to write DPT 16 to bus. I tested it and it works in my envrionment.


thelsing:
Merged. Thanks.

PR #30, opened 2019-08-25
added ESP32 platform

metaneutrons:
This code adds support for the ESP32 platform. It's very similar to ESP8266 but needed some changes in the multicast code.


thelsing:
Merged. Thanks for your work.

PR #35, opened 2019-09-13
Progbutton interrupt

mumpf:
Hi Thomas,

this allows to set the programmig button to generate interrupts on a FALLING signal, the default remains RISING. Needed for Sensormodule from @masifi, where the progbutton is buildin, but differntly wired.

Sorry for all the whitespace deltas, but it seems that our editors handle whitespaces differently...

Regards, Waldemar


thelsing:
Merged. Thanks.

PR #38, opened 2019-10-08
"Restart device" implementation

mumpf:
Hi, this is the implementation of the restart command. I tested it with many devices, it works as expected. It would be great if you can add this to the stack.


thelsing:
Thanks. I will look at this this weekend.

thelsing:
Can you please convert the different number for _restartState to an enum? I don't like magic numbers.

An other improvement that can wait for the next knx-command that needs a connection:
I think the application layer should establish a connection if needed. All telegrams that need to be buffered should be queued. After the connection is established, all the queued telegrams are sent and the corresponding callbacks are called. 
Feel free to implement this if you want. But it is not necessary for this pull request.

mumpf:
Hi,

I have changed all magic numbers to an enum. 

The other improvement is on my list, but I don't see a chance to implement it now.

thelsing:
merged thanks

PR #42, opened 2019-10-27
Add new KNX medium type RF for SAMD and Linux platform

nanosonde:
Please see README.md in the examples folder for more details.

RF for Linux is basically working, but not very well at the moment. Reading and writing of GPIOs on Linux has to be optimized with respect to speed (tested on Raspberry Pi). Also noise interference from the Raspberry Pi HW seems to be much higher in my setup (HW setup needs to be improved). RX sensitivity is not very well, because of the HW setup with loose wires.

RF for SAMD21 (Arduino) works pretty well as this platform had the focus.


src/knx/cemi_frame.h
```diff
@@ -62,4 +75,9 @@ class CemiFrame
     TPDU _tpdu;
     APDU _apdu;
     uint16_t _length = 0; // only set if created from byte array
+
+    // nly for RF medium
+    uint8_t* _rfSerialOrDoA = 0;
```
thelsing: Are they encoded in the data somewhere or is this some kind of additional info not included in the actual cemi-frame?

src/knx/data_link_layer.cpp
```diff
@@ -49,12 +54,18 @@ void DataLinkLayer::frameRecieved(CemiFrame& frame)
...
     if (source == ownAddr)
         _deviceObject.induvidualAddressDuplication(true);
 
     if (addrType == GroupAddress && destination == 0)
-        _networkLayer.systemBroadcastIndication(ack, type, npdu, priority, source);
+    {
+        if (systemBroadcast == SysBroadcast)
+            _networkLayer.systemBroadcastIndication(ack, type, npdu, priority, source);
+        else 
+            _networkLayer.dataIndication(ack, addrType, destination, type, npdu, priority, source);
```
thelsing: shouldn't this be be broadcatIndication from p. 9 of 3/3/3?

src/knx/device_object.h
```diff
@@ -44,8 +50,11 @@ class DeviceObject: public InterfaceObject
     uint8_t _prgMode = 0;
     uint16_t _ownAddress = 0;
     uint16_t _manufacturerId = 0xfa; //Default to KNXA
-    uint32_t _bauNumber = 0;
+    uint32_t _bauNumber = 0xaabbccdd;
```
thelsing: Is there a reason why this can't stay zero on default?

src/knx/platform.h
```diff
@@ -29,6 +29,15 @@ class Platform
     virtual int readUart() = 0;
     virtual size_t readBytesUart(uint8_t* buffer, size_t length) = 0;
 
+    virtual void setupSpi() = 0;
+    virtual void closeSpi() = 0;
+    virtual int readWriteSpi (uint8_t *data, size_t len) = 0;
+
+    virtual void setupGpio(uint32_t dwPin, uint32_t dwMode) = 0;
```
thelsing: See comment on arduino_platform.

src/linux_platform.cpp
```diff
@@ -519,4 +592,326 @@ void LinuxPlatform::cmdLineArgs(int argc, char** argv)
     memcpy(_args, argv, argc * sizeof(char*));
     _args[argc] = 0;
 }
-#endif
\ No newline at end of file
+
+void LinuxPlatform::setupGpio(uint32_t dwPin, uint32_t dwMode)
```
thelsing: Maybe wiringpi or something similar is a bit faster? (You mentioned timing problems in the readme.)

knx-linux/CMakeLists.txt
```diff
@@ -38,3 +77,5 @@ target_link_libraries(knx-linux  "${LIBRARIES_FROM_REFERENCES}")
 include_directories(../src)
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -Wall -Wno-unknown-pragmas -Wno-switch -g -O0")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Wall -Wno-unknown-pragmas -Wno-switch -g -O0")
+set_property(TARGET knx-linux PROPERTY CXX_STANDARD 11)
+install(TARGETS knx-linux RUNTIME DESTINATION /tmp)
```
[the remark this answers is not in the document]
nanosonde: I have just added this so that Qt Creator (my IDE) "knows" where to find the binary for deploying to the target while developing on my x86 Linux host machine. As I am running this as a normal user the usual paths /usr/local/bin, etc. won't work. So it is really a temporary file only while developing. Do you have any idea how to solve this?
BTW: Qt creator is also the reason why I have added all header files so that they show up in the project tree.

examples/knx-rf-demo/knx-rf-demo.ino
```diff
@@ -0,0 +1,70 @@
+#include <knx.h>
```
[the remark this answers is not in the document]
nanosonde: Not really. I thought that there might come RF specific stuff like semi-directional devices which are bidirectional while be configured and unidirectional during runtime.
Shall we remove this folder and keep it with the normal knx-demo.ino for now?

src/arduino_platform.cpp
```diff
@@ -137,6 +139,44 @@ size_t ArduinoPlatform::readBytesUart(uint8_t *buffer, size_t length)
...
+    SPI.end();
+}
+
+int ArduinoPlatform::readWriteSpi(uint8_t *data, size_t len)
+{
+    SPI.transfer(data, len);
+    return 0;
+}
+
+void ArduinoPlatform::setupGpio(uint32_t dwPin, uint32_t dwMode)
```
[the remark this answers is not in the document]
nanosonde: Yes, I have realized it and to be honest that was one of the most struggling part for me. ;)
i will remove the GPIO stuff from the platform base class again and provide global symbols for pinmode() etc.

src/knx/bau_systemB.cpp
```diff
@@ -350,3 +350,36 @@ void BauSystemB::nextRestartState()
...
+{
+    uint8_t knxSerialNumber[6];
+    uint8_t operand;
+
+    popByte(operand, testInfo + 1); // First byte (+ 0) contains only 4 reserved bits (0)
+
+    // See KNX spec. 3.5.2 p.33 (Management Procedures: Procedures with A_SystemNetworkParameter_Read)
+    switch(operand)
+    {
+        case 0x01: // NM_Read_SerialNumber_By_ProgrammingMode
```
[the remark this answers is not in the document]
nanosonde: ok.

src/knx/cemi_frame.h
nanosonde: You are right. This is contained in the RF additional info. 
However, it does not make sense to artificially create an RF info additonal info structure and add it to the buffer just to let it parse again in the cEMI frame class. The specification also says that for local usage in the stack it is not necessary to use additonal info. It is really only meant for messages which are leaving one device and enter another one (USB RF stick for example).
So I would like keep it there as is.
Should someone implement the USB HID KNX data link layer, the cEMI class would need refactoring anyway so that it is able to parse additional infos.

src/knx/data_link_layer.cpp
nanosonde: Yes. But as you have already pointed out the primitives were missing. 
I can add them accordingly. It is cleaner then.

src/knx/device_object.h
nanosonde: Yes, the reason was debugging. Have zero everywhere sometimes hides the fact that not enough bytes were copied for example. Easpecially when included in payload of an APDU, so at the end.
I can remove it again.

src/knx/platform.h
nanosonde: ok. Will change this.

src/linux_platform.cpp
nanosonde: I thought about WiringPI. Then I came accross this bad news: http://wiringpi.com/wiringpi-deprecated/
Bottom line: the author is fed up.

So should we still go with this? 
WiringPi offers both: direct HW access and sysfs.

For GPIO and SPI I use the standard kernel interface without any additional library at the moment. The SPI stuff here was inspired by WiringPi anyway. ;)
The current sysfs stuff is currently not used in proper way. I always open/close the files which is horribly slow. So maybe by keeping it open, it may work a low better.

Using the standard kernel interfaces would make it easy to use other boards (not RPi) as well.

But maybe it is really better to go with WiringPi although the orignal author has abandonned it. I am pretty sure that someone will take over this.

What do you think?

nanosonde: Maybe we should leave the Linux platform as generic one as is for KNX IP only and introduce a new platform specifically for Raspberry Pi.
Then I could implement the direct GPIO HW access there.
For SPI I would use the SPI kernel interface as I do now.

nanosonde: Maybe we should leave the Linux platform as generic one as is for KNX IP only and introduce a new platform specifically for Raspberry Pi.
Then I could implement the direct GPIO HW access there.
For SPI I would use the SPI kernel interface as I do now.

nanosonde:
Would it be ok for you if I leave out the Linux platform support at the moment to get this PR merged?
It would only support Arduino (SAMD21) then.

knx-linux/CMakeLists.txt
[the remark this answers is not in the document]
thelsing: I don't use Qt Creator, but for commandline you can set INSTALL_PREFIX or DESTDIR to change the prefix from /usr/local to something else like "/home/foobar". Qt Creator should have some option to set this.

examples/knx-rf-demo/knx-rf-demo.ino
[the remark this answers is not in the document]
thelsing: Yes please.

src/knx/bau_systemB.cpp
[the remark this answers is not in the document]
thelsing: Just keep it in bau_systemB.cpp for now. It can be move somewhere else if necessary.

src/knx/cemi_frame.h
thelsing: Ok.

src/knx/device_object.h
thelsing: Please do.

src/knx/data_link_layer.cpp
thelsing: Please do so.

src/linux_platform.cpp
thelsing: I think standard linux interfaces would be better. So the actual linux board doesn't matter. 
You can keep it as it is. Maybe you or I or somebody else will improve it later.

thelsing:
I think the linux support doesn't stop the pull request. Just convert the platform.gpio* method to the default arduino functions and leave the implementation as it is in linux_platform.gpio*. The implementation of the gpio* functions can be improved later.

src/arduino_platform.cpp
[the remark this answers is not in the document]
thelsing: Thanks.

thelsing:
Please set a default for MEDIUM_TYPE (to TP would be best). So the build will work again.

thelsing:
Can you please add #ifs to the RF-specific files depending on MEDIUM_TYPE == RF or something to fix the travis build?

examples/knx-rf-demo/knx-rf-demo.ino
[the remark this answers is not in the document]
nanosonde: done.

src/knx/bau_systemB.cpp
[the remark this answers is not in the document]
nanosonde: done.

src/knx/device_object.h
nanosonde: done.

src/arduino_platform.cpp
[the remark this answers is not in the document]
nanosonde: done.

src/knx/platform.h
nanosonde: done.

src/linux_platform.cpp
nanosonde: done.

src/knx/cemi_frame.h
nanosonde: nothing done.

src/knx/data_link_layer.cpp
nanosonde: done.

PR #45, opened 2019-11-23
Add support for Nucleo32 board (STM32 Cortex-M microcontroler)

condo4:
This add support for the board Nucleo32-L432KC, and is able to be quickly ported to all STM32 micro controleur using ST CubeMX projects.
A full example with CMake is available on
http://github.com/condo4/knx-pt100


nanosonde:
@condo4 
I would prefer it if you would also include your whole project ("knx-pt100") as a dedeciated folder within this repo. This way it would serve the purpose of being an example how to actually use the STM32 platform. Otherwise your addition would be of not much value, if you once decide to remove your GitHub project which contains all the essential code that is required to get the STM32 platform running. Just my 2 cents...

thelsing:
Thank you for the pull request. Can you please put all ring-buffer stuff into ringbuffer.c and ringbuffer.h? Also can you please refactor the code, so that the declarations and definitions common to non-arduino platforms is shared and not duplicated between them?

src/stm32_platform.cpp
```diff
@@ -0,0 +1,416 @@
+#include "stm32_platform.h"
+#include <stdio.h>
+#include <errno.h>
+#include <sys/stat.h>
+#include <sys/times.h>
+#include <sys/unistd.h>
+#include <knx/bits.h>
+#include <string.h>
+#include <ring_buffer.h>
+#include <main.h>
```
thelsing: where is main.h?

src/stm32_platform.cpp
```diff
@@ -0,0 +1,416 @@
+#include "stm32_platform.h"
+#include <stdio.h>
+#include <errno.h>
+#include <sys/stat.h>
+#include <sys/times.h>
+#include <sys/unistd.h>
+#include <knx/bits.h>
```
thelsing: should be include ""

src/stm32_platform.cpp
```diff
@@ -0,0 +1,416 @@
...
+#include <string.h>
+#include <ring_buffer.h>
```
thelsing: should be include ""

src/stdio_print.cpp
```diff
@@ -0,0 +1,170 @@
+#include <stdio.h>
+#include <knx/bits.h>
```
thelsing: should be include ""

src/knx/data_link_layer.cpp
```diff
@@ -59,7 +60,43 @@ void DataLinkLayer::frameRecieved(CemiFrame& frame)
     NPDU& npdu = frame.npdu();
     uint16_t ownAddr = _deviceObject.induvidualAddress();
     SystemBroadcast systemBroadcast = frame.systemBroadcast();
-    
+
+#ifdef DEBUG_DATA_LINK_LAYER
```
thelsing: can you add this print stuff to a method of CemiFrame please?

knx.cmake
```diff
@@ -0,0 +1,44 @@
+message(STATUS "KNX Stack")
+
+include_directories("${CMAKE_CURRENT_LIST_DIR}/src")
```
thelsing: this shouldn't be necessary if you correct the includes below.

condo4:
I think it's not really a good idea.

In fact, I think the logic is to separate the "stack" to all specific 
projects...

(this is why I used git submodules).

The content of mcu-core is automatically generated by CubeMX (I don't 
wrote anything in this folder), and depend on the version of cubemx and 
st libs...

The only one file I wrote really is knx-pt100.cpp, but this file is 
totally linked to my specific application.


The think I need to add, is a documentation to how to create and start a 
new project with CubeMX and knx-stack.

Fabien


Le 24/11/2019 à 09:09, nanosonde a écrit :

nanosonde:
@condo4 
I understand your point. 

It would be best IMHO if you could then add a dir which contains some minimal template and description on how to actually use STM32 platform with this stack.

All other MCU platform use the Arduino compatible cores (ESP, ESP32, SAMD21). Is there any reason why you have not chosen to use STM32duino?

condo4:
In fact; STM32duino don't support each stm32 core, and for build a real 
project, it's not very efficient; you can't use all the stm32 possibilities.

To develop on stm32, the majority of projects used

- HAL with CubeMX

- MBed

- OpenCM3

I started with opencm3; but it's harder than using CubeMX.

CubeMX it easy to use, and you have quickly a working project.

But, it's initially a C project, and the main difficulty is to integrate 
C with C++ for KNX stack.

I planed to create a full C stack inspired by this arduino KNX stack in 
order to be simpler, smaller, and more efficient on this type of MCU.

But for this moment, I used this C++ implementation to progress.


Fabien


Le 26/11/2019 à 12:33, nanosonde a écrit :

condo4:
Hi,


Le 25/11/2019 à 15:22, thelsing a écrit :
>
> *@thelsing* commented on this pull request.
>
> Can you please make the existing linux-cmake file use the new knx.cmake?
>
Sorry, I don't understand, where is linux-cmake file ?


> And please don't add more features to this pull request. Please create 
> different pull-requests for them.
>
> Regards,
> Thomas
>
> ------------------------------------------------------------------------
>
> In src/stm32_platform.cpp 
> <https://github.com/thelsing/knx/pull/45#discussion_r350207122>:
>
> > @@ -0,0 +1,416 @@
> +#include "stm32_platform.h"
> +#include <stdio.h>
> +#include <errno.h>
> +#include <sys/stat.h>
> +#include <sys/times.h>
> +#include <sys/unistd.h>
> +#include <knx/bits.h>
> +#include <string.h>
> +#include <ring_buffer.h>
> +#include <main.h>
>
> where is main.h?
>
main.h is a file generated by the CubeMX project (in the mcu-core part 
in the parent project).

This file is not generic, it's different for each project will use this 
stack.


> ------------------------------------------------------------------------
>
> In src/stm32_platform.cpp 
> <https://github.com/thelsing/knx/pull/45#discussion_r350207298>:
>
> > @@ -0,0 +1,416 @@
> +#include "stm32_platform.h"
> +#include <stdio.h>
> +#include <errno.h>
> +#include <sys/stat.h>
> +#include <sys/times.h>
> +#include <sys/unistd.h>
> +#include <knx/bits.h>
>
> should be include ""
>
> ------------------------------------------------------------------------
>
> In src/stm32_platform.cpp 
> <https://github.com/thelsing/knx/pull/45#discussion_r350207405>:
>
> > @@ -0,0 +1,416 @@
> +#include "stm32_platform.h"
> +#include <stdio.h>
> +#include <errno.h>
> +#include <sys/stat.h>
> +#include <sys/times.h>
> +#include <sys/unistd.h>
> +#include <knx/bits.h>
> +#include <string.h>
> +#include <ring_buffer.h>
>
> should be include ""
>
> ------------------------------------------------------------------------
>
> In src/stdio_print.cpp 
> <https://github.com/thelsing/knx/pull/45#discussion_r350207579>:
>
> > @@ -0,0 +1,170 @@
> +#include <stdio.h>
> +#include <knx/bits.h>
>
> should be include ""
>
> ------------------------------------------------------------------------
>
> In src/knx/data_link_layer.cpp 
> <https://github.com/thelsing/knx/pull/45#discussion_r350208638>:
>
> > @@ -59,7 +60,43 @@ void DataLinkLayer::frameRecieved(CemiFrame& frame)
>       NPDU& npdu = frame.npdu();
>       uint16_t ownAddr = _deviceObject.induvidualAddress();
>       SystemBroadcast systemBroadcast = frame.systemBroadcast();
> -
> +
> +#ifdef DEBUG_DATA_LINK_LAYER
>
> can you add this print stuff to a method of CemiFrame please?
>
> ------------------------------------------------------------------------
>
> In knx.cmake 
> <https://github.com/thelsing/knx/pull/45#discussion_r350210114>:
>
> > @@ -0,0 +1,44 @@
> +message(STATUS "KNX Stack")
> +
> +include_directories("${CMAKE_CURRENT_LIST_DIR}/src")
>
> this shouldn't be necessary if you correct the includes below.
>
Yes, I'm agree for stm32_platform.cpp, but knx.h and knx/dpt.h must also 
be included by the application project itself (knx-pt100.cpp in my example).

So, to be independent to the folder name of the knx-stack submodule, 
used include_directories directive is the best solution.

thelsing:
Just leave the cmake file as it is for now. The duplicate declarations and the debug code is more important. A minimal example would be nice too. (Like the knx-demo arduino example.)

nanosonde:
@condo4 
Would it be possible for you to use platformio as build platform?
https://docs.platformio.org/en/latest/frameworks/stm32cube.html#framework-stm32cube
https://docs.platformio.org/en/latest/boards/ststm32/nucleo_l432kc.html#board-ststm32-nucleo-l432kc

It supports STM32Cube as a framework and it also the supports your board.

nanosonde:
@condo4 
Maybe you find this useful: https://community.platformio.org/t/using-stm32cubemx-and-platformio/2611/2

condo4:
Hi,

I didn't know platformio before, it's seem to be an interesting project;

I will check if I can use it.


Le 16/12/2019 à 17:26, nanosonde a écrit :

Issue #44, opened 2019-11-10
Possible issues with interface object server


nanosonde:
Hi!

While implementing the new cEMI server feature I have analyzed the read/write access of properties.

When using Net'n'node I found out that it does the following:
1) M_PropWrite.req: cEMI Server Object, PID_COMM_MODE set 0x00 (data link layer)
2) M_PropRead.req:  cEMI Server Object, PID_COMM_MODE to verify if it is really set to 0x00.

```
cEMI USB RX len: 7 data:  data: FC 00 08 01 34 10 01
M_PropRead_req
ObjectType: 0x0008
ObjectInstance: 0x01
PropertyId: 0x34 (52) -> PID_COMM_MODE (PDT_ENUM8)
NumberOfElements: 16
StartIndex: 1
```
However, the property is only PDT_ENUM8, so a single byte.
The client (Net'n'Node) requested 16 elements (is this a bug?).
So a valid response would now contain NumberOfElements=1, StartIndex=1 and a single data byte with the comm mode. 
(See KNX spec. 3/6/3 p.100: M_PropRead.con message, positive response)

I do not see this kind of implementation anywhere in an interface object implementation.
Am I overlooking something or did I find an issue here?
Current implementation always sends back the NumberOfElements received in the request, but not the actual available MaxNoOfElements. 
Also whenever an external client (A_PropRead) wants to read the property value length, the code won't work as the current code would handle a startIndex=0 to read the length entry (first array entry) of the property value.

See KNX spec. 3/4/1 p.15 that a property value of a full interface object (what you have implemented) shall be always(!) an array. First entry is the length, followed by the data.

nanosonde:
Argh. I have read the docs again.

NumberOfElements: 4 bits (and not 8 bits!)
StartIndex: 12 bits (and not 8 bits!)

Very badly illustrated in the docs. ;)

So it does not request 16 elements, but just 1 element which is correct for PDT_ENUM8 type.

However, the thing "property value is always an array with first entry being the length" remains an issue IMHO. This won't work with the current implementation.

thelsing:
You are right. PropertyValueRead already seems to do the right thing. 
Returning the right number of elements for PropertyWrite should be easy. 
Resetting the number of elements to no elements as written in 
4.3.2.2 conflicts a bit with 4.3.1.3  (both in 3/4/1) 
[...]

PR #47, opened 2019-12-05
Add cEMI Server with KNX USB HID support

nanosonde:
This PR adds an optional cEMI Server to the stack with KNX USB HID support.
* see further details in file knx_cemi_notes.md
* ETS and Net'N'Node recognize it as a normal KNX USB Data Interface
* KNX TP (untested) and RF (tested) supported
* Adafruit SAMD Arduino Core required (they include TinyUSB stack instead of the default Arduino USB stack. The latter does not offer all required features, see https://github.com/adafruit/ArduinoCore-samd/tree/master/cores/arduino/Adafruit_TinyUSB_Core)
  * The Adafruit core is automatically chosen on PlatformIO if the board is "adafruit_feather_m0".
  * Should work with normal Arduino IDE as well (see https://learn.adafruit.com/adafruit-feather-m0-basic-proto/using-with-arduino-ide)
* bau_systemB class had to be extended with an additional method to fetch object references as the tunnel local management interface uses parameters "ObjType" and "ObjInstance" instead of an index to access objects. Therefore all derived classes had to be touched.
* cEMI Server support is completely optional. See: #define USE_CEMI_SERVER


thelsing:
Thank you for the pull request. You changed quite some files, so it will take some time to read through this. Please be patient.

src/knx/application_program_object.cpp
```diff
@@ -25,7 +25,7 @@ void ApplicationProgramObject::readProperty(PropertyID id, uint32_t start, uint3
     }
 }
 
-void ApplicationProgramObject::writeProperty(PropertyID id, uint8_t start, uint8_t* data, uint8_t count)
+void ApplicationProgramObject::writeProperty(PropertyID id, uint32_t start, uint8_t* data, uint32_t& count)
```
thelsing: What is the reason for changing the signature? Are the uint8_t parameters not big enough?

src/knx/usb_tunnel_interface.cpp
```diff
@@ -0,0 +1,543 @@
...
+					respData[2] = 0x04; // USB KNX Transfer Protocol Body: Feature Data -> only cEMI supported
+					sendKnxHidReport(BusAccessServer, DeviceFeatureResponse, respData, 3);
+					break;
+				case HostDeviceDescriptorType0: // Host Device Descriptor Type 0
+					Serial1.println("Device Feature Get: Host Device Descriptor Type 0");
+					pushWord(_maskVersion, &respData[1]); // USB KNX Transfer Protocol Body: Feature Data -> Mask version
+					sendKnxHidReport(BusAccessServer, DeviceFeatureResponse, respData, 3);
+					break;
+				case BusConnectionStatus: // Bus connection status
+					Serial1.println("Device Feature Get: Bus connection status");
```
thelsing: The Serial1.println should probably be only println().

nanosonde: yes. fixed.

src/knx/application_program_object.cpp
nanosonde: uint8_t is not enough as startIndex has 12 bits length for A_PropertyValue_* and M_Prop_* services. so uint16_t could be enough for "start".

Concerning "count" I am unsure.
In 3/4/1 Application Interface Layer on page 15 (4.3.1.3 Property Value) it says that every property of a full interface object is always an array. Furthermore, array index 0 is the length of this array. BUT: the size of array index 0 can be either uint16_t or uint32_t depending of the value of "max_no_of_elements" in the property description. If the 4 bit binary exponent is 0, then it is uint16_t, otherwise it is uint32_t.
While writing this, I think that I have mixed up "count" which is basically the argument "nr_of_elements" in the  A_PropertyValue_* and M_Prop_* services and array index 0 which always contains the length of the array: either uint16_t or uint32_t.

Summary for A_PropertyValue_* and M_Prop_* services: 
* "start" (a.k.a. "startIndex in the spec, 12 bits) should be uint16_t
* count (a.k.a. "nr_of_elements, 4 bits) should be uint8_t

However, the interface object server implementation is still broken as it does not handle read/write requests to array index 0 properly as discussed here: https://github.com/thelsing/knx/issues/44

nanosonde: Mhmm... after thinking about this again, I think that "count" is just the number of bytes returned in "data". and not "nr_of_elements".

count = nr_of_elements * PropertySize

What is the maximum property size that is defined?

nanosonde: See table on page 18. PDT_GENERIC20 has 20 octets:
nr_of_elements (4 bits): 15 * 20octets = 300 octets 
So minimum uint16_t is required.

src/knx/usb_tunnel_interface.cpp
thelsing: Ok. I will change it after I finish debugging the new memory stuff.

thelsing:
Merged. Thanks.

PR #48, opened 2019-12-16
Travis knx usb

nanosonde:
Changed Travis setup to use matrix for each build:
* linux platform (cpp)
* arduino platform (cpp)
* platformio projects (python-2.7)

So I have removed the global env var "MODE" and put everything into its own matrix build entry.
This was needed as I had to include "language" also inside the build matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,16 +83,11 @@
         directories:
           - "~/.platformio"
       env:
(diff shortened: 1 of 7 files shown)


thelsing:
Merged. Thanks.

PR #49, opened 2019-12-17
Improve CI

nanosonde:
Added platformio build jobs for example knx-demo for :
* SAMD
* ESP8266
* ESP32


thelsing:
Merged. Thanks.

Issue #60, opened 2020-05-29
Properly handle SystemBroadcast communication on closed media


nanosonde:
Currently, application layer services like A_SystemNetworkParameterRead are only reachable via SystemBroadcast. 
However, on closed media there is no additional SystemBroadcast, instead only normal Broadcast mode is used.
(note: a coupler between closed and open media has to properly handle this situation by changing the mode accordingly -> change the SBC flag)

See Application Layer PDF p.30.

Basically the content of [this method](https://github.com/thelsing/knx/blob/cc1fc7d5fb5050309937e99b9e0b346f2132292f/src/knx/application_layer.cpp#L167) must also be evaluated on closed media even if it is on broadcast only.

nanosonde:
I close this for now, as coupler support is not implemented yet.

nanosonde:
Just saw this entry:
https://support.knx.org/hc/de/articles/360009162540-ETS-v5-7-3

KNX Secure

    Support für IP-SystemBroadcast in Netzwerkmanagementverfahren für $57B0

PR #62, opened 2020-06-08
Refactor broadcast/systembroadcast

nanosonde:
Get rid off redundant code and introduce addtional argument (type of broadcast) to method calls.

Also fixes this issue: https://github.com/thelsing/knx/issues/60

### DIFF
--- a/src/knx/application_layer.cpp
+++ b/src/knx/application_layer.cpp
@@ -239,20 +239,20 @@ void ApplicationLayer::dataBroadcastConfirm(AckType ack, HopCountType hopType, P
             popByte(testInfo[1], data + 5);
             propertyId = (propertyId >> 4) & 0x0FFF;;
             testInfo[0] &= 0x0F;
-            //TODO: _bau.systemNetworkParameterReadLocalConfirm(priority, hopType, objectType, propertyId, testInfo, sizeof(testInfo), status);
+            _bau.systemNetworkParameterReadLocalConfirm(priority, hopType, objectType, propertyId, testInfo, sizeof(testInfo), status);
             break;
         }
         case DomainAddressSerialNumberWrite:
         {
             const uint8_t* knxSerialNumber = &data[1];
             const uint8_t* domainAddress = &data[7];
-            //TODO: _bau.domainAddressSerialNumberWriteLocalConfirm(priority, hopType, domainAddress, knxSerialNumber, status);
+            _bau.domainAddressSerialNumberWriteLocalConfirm(priority, hopType, domainAddress, knxSerialNumber, status);
             break;
         }
         case DomainAddressSerialNumberRead:
         {
             const uint8_t* knxSerialNumber = &data[1];
-            //TODO: _bau.domainAddressSerialNumberReadLocalConfirm(priority, hopType, knxSerialNumber, status);
+            _bau.domainAddressSerialNumberReadLocalConfirm(priority, hopType, knxSerialNumber, status);
             break;
         }
         default:
--- a/src/knx/application_layer.cpp
+++ b/src/knx/application_layer.cpp
@@ -118,8 +118,15 @@ void ApplicationLayer::dataBroadcastIndication(HopCountType hopType, Priority pr
             break;
         }
         default:
+#if (MEDIUM_TYPE == 5)||(MEDIUM_TYPE == 0)
+            // Make sure we also check if it is a service normally available only on SystemBroadcast on open media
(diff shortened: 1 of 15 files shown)


src/knx/data_link_layer.cpp
```diff
@@ -41,11 +41,16 @@ void DataLinkLayer::dataRequest(AckType ack, AddressType addrType, uint16_t dest
...
 }
 
-void DataLinkLayer::systemBroadcastRequest(AckType ack, FrameFormat format, Priority priority, NPDU& npdu)
+void DataLinkLayer::broadcastRequest(AckType ack, FrameFormat format, Priority priority, NPDU& npdu, SystemBroadcast broadcastType)
 {
     // System Broadcast requests will always be transmitted as broadcast with KNX serial number for open media (e.g. RF medium) 
     // See 3.2.5 p.22
-    sendTelegram(npdu, ack, 0, GroupAddress, format, priority, SysBroadcast);
+#if (MEDIUM_TYPE == 5)||(MEDIUM_TYPE == 0)
+    (void)broadcastType; // not used on TP/IP, always normal broadcast
```
thelsing, line 49: is the (void)broadcastType; debug code or necessary?

src/knx/application_layer.cpp
```diff
@@ -181,32 +239,36 @@ void ApplicationLayer::dataSystemBroadcastIndication(HopCountType hopType, Prior
             popByte(testInfo[1], data + 5);
             propertyId = (propertyId >> 4) & 0x0FFF;;
             testInfo[0] &= 0x0F;
-            _bau.systemNetworkParameterReadIndication(priority, hopType, objectType, propertyId, testInfo, sizeof(testInfo));
+            //TODO: _bau.systemNetworkParameterReadLocalConfirm(priority, hopType, objectType, propertyId, testInfo, sizeof(testInfo), status);
```
thelsing, line 242: Why are these calls commented out? They should have an empty standard implementation, shouldn't they?

src/knx/data_link_layer.cpp
nanosonde: I was to lazy to put __attribute__((unused)) in these cases. Shall I use it?
It suppresses compiler warnings about unused vars.

src/knx/application_layer.cpp
nanosonde: As they are not used I did not implement the empty ones. Shall I do implement the empty stuff?

nanosonde: BTW: you cannot see it really well from the DIFF.
Have a look here: https://github.com/nanosonde/knx/blob/refactor_broadcast/src/knx/application_layer.cpp#L80
And here: https://github.com/nanosonde/knx/blob/refactor_broadcast/src/knx/application_layer.cpp#L179

nanosonde:
I just unified them to save code and to have only a single place to handle both broadcast types.
For RF and PL, some AL services are only reachable via SystemBroadcast. In contrast on TP and IP all AL services shall be reachable via Broadcast only.
For a coupler BAU this will become more important.
The coupler needs to translate between SystemBroadcast and Broadcast by looking at the APCI type in the network layer and route the message accordingly when routing between a closed and an open medium.
But even for the coupler bau with primary and secondary data link layers, it would be absolutely sufficient to have this unification and just look at the flag.

src/knx/application_layer.cpp
nanosonde: Just added the empty confirm counterparts.

src/knx/data_link_layer.cpp
nanosonde: I have removed the line for now.
I just realized that there are tons of warnings about unused vars everywhere else. :)
So maybe we should silence this warning for now as it really disturbs the compilation log essentials.

nanosonde:
Mhmm. Thinking about it, you are right: it is closer to the specs, 
Shell we keep the seperate splitted broadcast methods?
I could just change the AppLayer code to make sure that SystemBroadcast AL service are also available for IP and TP which only use Broadcast.

thelsing:
Then let's keep them separated and only change the application layer.

nanosonde:
@thelsing Please review again.

nanosonde:
@thelsing forgot the sending direction. Now everything should be in place.

thelsing:
Seems fine to me. If we create couplers in the future we will have to change the #ifdefs to something else.

thelsing:
Merged. Thanks.

PR #65, opened 2020-06-08
Disable Travis-CI 8266 Arduino test

nanosonde:


nanosonde:
At least the other Arduino sketches are tested against SAMD2 platform.

thelsing:
The whole CI should be migrated to platform.io I think.

PR #71, opened 2020-06-27
Stm32 port

etrinh:
Fix with unitialized member
+ Few minor fixes
Hardware: STM32L432KC + NCN5130


thelsing:
Can you change the tab indent to spaces please? Otherwise this is fine.

src/knx/config.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#ifndef NO_KNX_CONFIG
```
thelsing, line 3: What is this for? I think this got mixed in?

etrinh: Right
This was to avoid to modify files inside library tree and keep all externals always vanilla.
Specific configurations should better be placed in settings defines.
I noticed a similar behaviour where platforms are linked with specific media.

thelsing:
Merged thanks.

PR #72, opened 2020-06-28
check chip state when enabled

leoujz:


src/knx/tpuart_data_link_layer.cpp
```diff
@@ -383,6 +387,28 @@ bool TpUartDataLinkLayer::sendFrame(CemiFrame& frame)
     return true;
 }
 
+void TpUartDataLinkLayer::stateChip() {
+    // check chip state if enabled and no data received more than 10 seconds
+    if (_enabled && millis() - _lastByteRxTime > 10000) {
+        uint8_t cmd = U_STATE_REQ;
+        _platform.writeUart(cmd);
+        _waitConfirmStartTime = millis();
+        while (true)
```
thelsing, line 396: The while loop here blocks the main loop. This is not good. Can you please refactor this in the same way the rest of the tpuart-layer works?

thelsing:
Ne reaction. Closing.

PR #75, opened 2020-07-03
Bugfixes for property startIndex 0 and PID_IO_LIST

nanosonde:
startIndex = 0 must return two bytes which contain the size of the property.

PID_IO_LIST: Due to missing explicit casting, the compiler chooses uint8_t variant of overloaded method.


nanosonde:
I have just looked through the other objects.
The readcallback of all callback properties have to be adjusted as well. 
While going through the object, I wondered if callback properties are really necessary in most cases. For example if in the IP object. Most properties could have also been realized as DataProperties, couldn‘t they?

nanosonde:
Ok, just tested this with Net'n' node. Querying of the objects and properties seems to work correctly now.
I have also fixed the table object class. Same issue with the callback properties.

You can merge it now. :wink:

thelsing:
Merged. Thanks.

PR #77, opened 2020-07-09
Try fixing a crash due to NULL derefence

etrinh:
With an ESP32 + TPUART configuration, I noticed a crash at program startup in the TpUartDataLinkLayer::loop()
It seems that the _sendBuffer can be NULL when used in logic.
I have naively added a logic to avoid this NULL dereference which works at my side, but I am not sure if this is correct with the whole project design, please don't hesitate to comment.
Thanks

### DIFF
--- a/src/knx/tpuart_data_link_layer.cpp
+++ b/src/knx/tpuart_data_link_layer.cpp
@@ -245,7 +245,7 @@ void TpUartDataLinkLayer::loop()
                 //Destination Address + payload available
                 _xorSum ^= rxByte;
                 //check if echo
-                if (!((buffer[0] ^ _sendBuffer[0]) & ~0x20) && !memcmp(buffer + _convert + 1, _sendBuffer + 1, 5))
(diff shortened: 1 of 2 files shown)


etrinh:
Add another tweak, because my ESP32 EEPROM was configured to 0x1000 so it wasn't able to handle the default 8192 and crashed. Make it possible to change it by defining 'KNX_FLASH_SIZE'.

etrinh:
I just checked the 8192 flash size on my STM32L432 it has even less (2048).

src/knx/tpuart_data_link_layer.cpp
```diff
@@ -245,7 +245,7 @@ void TpUartDataLinkLayer::loop()
                 //Destination Address + payload available
                 _xorSum ^= rxByte;
                 //check if echo
-                if (!((buffer[0] ^ _sendBuffer[0]) & ~0x20) && !memcmp(buffer + _convert + 1, _sendBuffer + 1, 5))
+                if (_sendBuffer != NULL && (!((buffer[0] ^ _sendBuffer[0]) & ~0x20) && !memcmp(buffer + _convert + 1, _sendBuffer + 1, 5)))
```
thelsing, line 248: can you change NULL to nullptr please?

Issue #76, opened 2020-07-07
Add coupler support


nanosonde:
### Possible couplers according to AN161 (coupler model 2.0) and coupler model 1.x.

IP/TP1 coupler: 091A - coupler model 1.x
TP1/RF coupler: 2920 - coupler model 2.0
TP1/TP1 coupler: 0912 - coupler model 1.x

### Structure:
Device BAUs: bau07b0, bau27b0, bau57b0 <- bau_systemB_device <- bau_systemB <- bau
Coupler BAUs: bau091a, bau2920, bau0912 <- bau_systemB_coupler <- bau_systemB <-bau

### Tasks:
- ~~Move group communication (with group objects) and all normal device related code to bau_systemB_device~~ DONE
- ~~Keep common code for interface object server access, restart, etc. in bau_systemB~~ DONE
- ~~Rename network_layer.* to network_layer_device.*~~ DONE
- ~~Create new network_layer_coupler.* which is instantiated in bau_systemB_coupler~~ DONE
- ~~Move some current existing routing relevant code (few lines only) from datalink_layer.* to network_layer~~ DONE
  - ~~comparision of dstAddr of incoming packet is part of the routing decision which is normally done in the network layer. The datalink_layer shall just forward **all** received frames accordingly (N_GROUP, N_DATA, N_BROADCAST, N_SYSTEMBROADCAST)~~ DONE
- ~~Instantiate two datalink_layers in bau_systemB_coupler~~ DONE
  - ~~primary side: example ip_datalink_layer~~ DONE
  - ~~secondary side: example tp_datalink_layer~~ DONE
- ~~New network_layer_coupler.* has connections to both datalink layers~~ DONE
  - ~~implement network layer entity class to which the usual data link layer code can connect~~ DONE
  - ~~provide two network layer entity instances for primary and secondary datalink layer connections~~ DONE
  - ~~implement routing between primary/secondary~~ DONE

### Open issues with current implementation
- ~~Application layer responses must be answered on interface from which the request was received~~ not an issue

nanosonde:
@thelsing 
I close this issue as inital coupler support was added.
Bugs/Improvements should be handled in there own tickets.

Maybe you could also add the label "feature" to this one.

PR #79, opened 2020-07-21
Add coupler support

nanosonde:
This PR adds coupler support. More details here: https://github.com/thelsing/knx/issues/76


examples/knx-linux-coupler/main.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+    uint8_t serialNumber[] = { 0x00, 0xFA, 0x01, 0x02, 0x03, 0x04};
+    uint8_t key[] = { 0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F};
+
+    FdskCalculator calc;
+    char fdskString[42]; // 6 * 6 chars + 5 dashes + nullbyte = 42
+    calc.snprintFdsk(fdskString, sizeof(fdskString), serialNumber, key);
+    printf("FDSK: %s\n", fdskString);
+
+    // Prevent swapping of this process
```
thelsing, line 78: Just out of interest: why is it necessary to prevent swapping?

src/knx/tpuart_data_link_layer.h
```diff
@@ -8,19 +8,26 @@
 
 #define MAX_KNX_TELEGRAM_SIZE 263
 
+class ITpUartCallBacks
```
thelsing, line 11: I would have use a function pointer for this. But I don't have a strong opinion about it. Not sure whats better. A function pointer may be a bit faster. Any opinion on this? I would like to have callbacks handled in a uniform way across the project.

examples/knx-linux-coupler/main.cpp
nanosonde: This was introduced when I added KNX-RF. As we only have the single main loop and RF timing is critical, I have added this to make sure that any process memory is swapped to disk.

src/knx/tpuart_data_link_layer.h
nanosonde: Normally I would have used std::function for this. I have already found some implementation which do not use dynamic memory for this.
Example: https://github.com/arobenko/embxx/blob/master/embxx/util/StaticFunction.h

nanosonde: This was my try: https://github.com/nanosonde/knx/blob/feat_coupler/src/knx/simple_functional.h
But I have decided not to use it as it seems to be a mess.

nanosonde:
Just a small note:
I have decided to NOT split up the TransportLayer and ApplicationLayer classes each of it into two parts: one for the management stuff and one for the group comm stuff as I did it for the bau_SystemB class.
Therefore I made group comm related things optional by using a pointer which could be nullptr.

src/knx/tpuart_data_link_layer.h
thelsing: So simple_functional is not used? Then it shouldn't be in the pull request (and I don't have to try to understand it ;)

thelsing:
You're to fast. I still have to look over 50 files or so.

examples/knx-linux-coupler/main.cpp
nanosonde: I have no proof that this really has any effect. As KNX-RF with a CC1101 directly connected via SPI is really ugly. The RPi causes so much interference the RX sensitivity is really bad. Ok, with a proper PCB layout and some filtering this could be handled, but I am not that much interested in using the CC1101 directly with the RPi. It served the purpose of being a dev platform to test things quickly related to the CC1101.

src/knx/tpuart_data_link_layer.h
nanosonde: Yes, sorry. I forgot to remove it.

nanosonde:
Yeah, ok. I will wait with my comments. :-)

src/knx/bau57B0.cpp
```diff
@@ -9,16 +9,17 @@
...
       _ipParameters(_deviceObj, platform),
-      _dlLayer(_deviceObj, _addrTable, _ipParameters, _netLayer, _platform)
+      _dlLayer(_deviceObj, _ipParameters, _netLayer.getEntity(0), _platform)
 #ifdef USE_CEMI_SERVER
       ,
       _cemiServer(*this)
 #endif
 {
-    _netLayer.dataLinkLayer(_dlLayer);
+    _netLayer.getEntity(0).dataLinkLayer(_dlLayer);
```
thelsing, line 20: Just to nitpick: One could eliminate the magic number 0 ;)

thelsing (approved):
Seems fine to me. Some comments:
- documentation: can you a bit documentation how to use it? Is it possible to use ESP8266 with TPUART as an IP/TP Router? It would be nice if you could update the class diagramm (Same for the secure stuff)
- you introduced the MASK-defines. We should use the in the baus to only compile the bau if its MASK is defined
- for further pull requests: It would've been helpful if you described you changes in a few sentences. 

Nothing to keep this from getting merged. Just nice to have. What do you say?

src/knx/bau57B0.cpp
thelsing: Same with all other getEntity calls. Are there more than 2 of them? If not getPrimary and getSecondary are easier to understand.

nanosonde: The current KNX spec. does not forsee more than two interfaces, although they mention in the coupler model 2.0 spec that it could be an option in the future.
I will add public members getPrimary() and getSecondary().

src/knx/tpuart_data_link_layer.h
nanosonde: Maybe we should checkout this implementation: https://github.com/arobenko/embxx/blob/master/embxx/util/StaticFunction.h
It would have to be adapted though to replace the GASSERT().

nanosonde: It will still require the header #include <functional>

nanosonde: Please also read this: https://arobenko.gitbooks.io/bare_metal_cpp/content/basic_needs/callback.html

thelsing: I'd rather keep it as it is. I don't really see why we should use a complicated template. The current solution is much easier to understand.

nanosonde: Ok. However, after thinking about it, I would say that the implementation should be changed.
It is the responsibilty of the network layer to give an answer if a packet should be ACK'ed or not.
For a router and also for normal device. So instead of placing it in the BAU, I think there should be an interface via the network entity to ask the network layer "shallPackedBeAcked". Then there is no need for registering a callback. It is then part of the normal network layer (entity) interface towards the datalink layer.
What do you think?

nanosonde: Note: If a datalink layer does not ACK packets then it simply does not ask the network layer. E.g.: IP and RF (more specific RF ready, RF multi can ACK packets)

thelsing: Feel free to change it.

src/knx/bau57B0.cpp
nanosonde: replaced.

src/knx/tpuart_data_link_layer.h
nanosonde: I will change this. Stay tuned.

nanosonde: After thinking about it again, I will leave it as is for now as it only affects TP data link layer which uses ACKs.

nanosonde:
Could you please check again?
I have added some minimal docs, updated the class diagram for normal devices and added class diagram for a IP/TP1 coupler.
Also changed the things you had requested except the callback for TPUART ACKs.

Travis CI also modified. ESP8266 commented out as I had problems with it: https://github.com/thelsing/knx/issues/86

thelsing:
Merged. Thanks as always.

PR #94, opened 2020-10-20
Add support for CC1310 platform based on SimpleLink SDK

nanosonde:
This PR adds support for CC1310 platform based on SimpleLink SDK.
The SDK is realized as a GIT submodule which points to a contiki-ng repository on GitHUB.
It is just a subset of the original TI SDK which only offers the version without RTOS.
No multihreading is used.
The initial ideas are from @mjm987 that I have incorporated.

CMake is used as a build system.
Be sure to have a recent arm-none-eabi toolchain installed. I am using the one that comes with Ubuntu-Mate 20.04.

My dev environment is VScode with [Cortex-Debug extension](https://marketplace.visualstudio.com/items?itemName=marus25.cortex-debug) and a Segger JLink EDU.
I also recommend to install the[ device support pack](https://marketplace.visualstudio.com/items?itemName=jeandudey.cortex-debug-dp-cc26xx-cc13xx) for this device family to have register view etc.

The onboard XDS110 debug interface is rather slow and only works properly with [TI CCS](https://www.ti.com/tool/CCSTUDIO) which is Eclipse based IDE that I do not prefer.
As a result I also added [Segger RTT](https://www.segger.com/products/debug-probes/j-link/technology/about-real-time-transfer/) which is really fast and does not occupy another UART.


nanosonde:
Not finished yet!

thelsing:
I would prefer a non submodule solution. I slightly remember that there are some other alternatives. I will look into this at the weekend.

nanosonde:
I could use this:

https://cmake.org/cmake/help/latest/module/FetchContent.html

thelsing:
I was thinking into the direction of git subtree. Se we don't depend on external repositories.

mjm987:
@nanosonde: 
Good work and very interesting but does it really run this way including cc310 RF? 
I assume the problem is still that your main() runs after the knx constructors which are called in startup_cc13xx_cc26xx_gcc.c by iterating over __init_array_start[] .
So I guess you:
- either has to call the init stuff in which you did in main() in the startup code _before_ calling the constuctors
- or you could define in main a "constructor function" as follows which is called as very first constructor function:

__attribute__((constructor)) void early_init_constructor(void)
{
    early_init();
}

mjm987:
As mentioned once the problem with TI-RTOS according init was that calling BIOS_start() does not return because it starts the scheduler and uses the main stack as exeption stack.
So if I did the early init in a constructor function including calling BIOS_start() at the end of it, the remaining constructors were not called anymore.
I solved this now by calling the remaining constructors at start of mainThread() by iterating over  __init_array_start[i] starting form i=1 :
```
extern void (*__init_array_start []) (void);
extern void (*__init_array_end []) (void);
int count = __init_array_end - __init_array_start;
for (int i = 1; i < count; i++) {
    __init_array_start[i]();
}
```

nanosonde:
@mjm987 
No constructors are called before main during startup.
See the map file:
```
*(.ctors)
                0x000000000001b7c8                . = ALIGN (0x4)
 *(.dtors)
                0x000000000001b7c8                . = ALIGN (0x4)
                0x000000000001b7c8                __init_array_start = .
 *(.init_array*)
                0x000000000001b7c8                __init_array_end = .
 *(.init)
 *(.fini*)
```

The init array is empty. Why do you think there should be constructors that are called during startup?
The current implementation here does NOT use any static initializers.
To put it simple: all KNX sub classes are basically members in one of the classes which have class platform and class bau as base class. So they are constructed (ctor is called) when a derived class vom platform or bau gets instantiated. This is the case in class knx_facade where both are instantiated with new.

In C++ it is an advise to not use static global instances of a class (hint: "static initialization order ‘fiasco’ (problem)").
See this FAQ here: https://isocpp.org/wiki/faq/ctors#static-init-order

The code here does NOT use any static global instances. Again: all KNX classes are somehow contained as member vars.

Another example:
```
class A
{
    public:
    A() : x {42} {}
    void foo() { x = 0; }
    
    private:
    int x;
};
```

This here (1): 
```
int main()
{
    A a;
    a.foo();
}
```

Is not the same as (2):
```
A a;
int main()
{
    a.foo();
}
```

In (1) class A is instantiated (ctor called) **after** jumping to main()
In (2) class A is instantiated (ctor called) **before** jumping to main(). Ctor is called during __libc_init_array.

nanosonde:
Small offtopic hint:  https://github.blog/changelog/2018-08-17-collapse-all-diffs-in-a-pull-request-at-once/ :smile:

mjm987:
ok sorry, I missed knx_wrapper.cpp ;-)
Does RF and standby modes work as well?

src/knx_facade.cpp
```diff
@@ -42,14 +42,19 @@
     #endif
 #elif __linux__
```
thelsing, line 43: just threat linux as all other non arduino platforms here. There is no reason di differentiate them, right?

src/knx_facade.cpp
```diff
@@ -42,14 +42,19 @@
...
+    // no predefined global instance
 #endif
 
 #ifndef ICACHE_RAM_ATTR
     #define ICACHE_RAM_ATTR
 #endif
 ICACHE_RAM_ATTR void buttonUp()
 {
-    #ifndef __linux__
+    #if !defined(__linux__) && (defined(ARDUINO_ARCH_STM32) || \
```
thelsing, line 54: Same here.

thelsing (approved):
Quite some code. But as only a few bits touch other stuff, just give me a ping when you think that this is ready.

src/knx_facade.cpp
nanosonde: changed it.

src/knx_facade.cpp
nanosonde: also here

src/knx/rf_physical_layer.h
```diff
@@ -182,6 +182,9 @@ extern void delayMicroseconds (unsigned int howLong);
 #define TX_ACTIVE 4
 #define TX_END 5
 
+// Calculate the real packet size out of the L-field of FT3 frame data. See KNX-RF spec. 3.2.5 Data Link Layer frame format
+#define PACKET_SIZE(lField) ((((lField - 10 /*size of first pkt*/))/16 + 2 /*CRC in first pkt */) * 2 /*to bytes*/ +lField + 1 /*size of len byte*/)
```
nanosonde, line 186: I have written some testcode which runs this MACRO against the other packetSize() implementation for all input values from 0 to 255. Both calculate the same packet size for each input value.
So I replaced the old implementeation packetSize() with the MACRO.

thelsing:
Merged thanks.

PR #95, opened 2020-10-22
Add float type for parameters & a few fixes

ajlajlajl:
Added float data type support in Parameters.
Also fixed a few things (in STM32 platform mostly)

1. one include was not relative, i got error cuz i had not set an include directory.
2. while checking platforms with #ifdef's, using defined() in #ifel's was forgotten.
3. and a misspelled file extension


src/knx/application_program_object.cpp
```diff
@@ -47,3 +47,26 @@ uint32_t ApplicationProgramObject::getInt(uint32_t addr)
...
 }
+
+double ApplicationProgramObject::getFloat(uint32_t addr, ParameterFloatEncodings encoding)
+{
+    uint8_t buffer[8];
+    switch (encoding)
+    {
+        case Float_Enc_DPT9:
+            memcpyInverted(buffer, TableObject::data() + addr, 2);
+            return Decode_DPT9_Float(*(uint16_t*)buffer);
```
thelsing, line 58: can't https://github.com/thelsing/knx/blob/edb9b6b52670be1d6e5a30f261f790a5d090a1d5/src/knx/dptconvert.cpp#L1716 or some other function from dptconvert.cpp be reused here?

ajlajlajl: Sure! my bad. I didn't see those functions. 
a question:
for IEEE754 Double (64bit), should I add "float64FromPayload" and "float64ToPayload" functions to "dptconvert.cpp" ? considering the fact that its not used in any DataPoint?

thelsing: Yes I think this is the best place to put them.

ajlajlajl:
ok I've changed it as you suggested. 
But I cant test it cuz recent commits is not working ( on STM32 at least ), no response is received from device.

thelsing:
merged thanks.

PR #108, opened 2020-11-28
Making USE_DATASECURE optional

rueckix:
Having USE_DATASECURE defined by default breaks installations that do not use KNX Secure. 

Reason
Due the the way "bool BauSystemBCoupler::configured()" is implemented, "knx.configured()" will always return "false"

Proposed Solution
Let the user define USE_DATASECURE via compiler flags when required.

### DIFF
--- a/examples/knx-cc1310/CC1310_LAUNCHXL_NoRTOS.lds
+++ b/examples/knx-cc1310/CC1310_LAUNCHXL_NoRTOS.lds
@@ -34,8 +34,8 @@
  *  Default Linker script for the Texas Instruments CC1310
  */
 
-STACKSIZE = 4096+1024;
-HEAPSIZE = 2048;
+STACKSIZE = 4096;
+HEAPSIZE = 4096;
 
 MEMORY
 {
--- a/examples/knx-cc1310/RTT/SEGGER_RTT_Conf.h
+++ b/examples/knx-cc1310/RTT/SEGGER_RTT_Conf.h
@@ -72,7 +72,7 @@ Revision: $Rev: 18601 $
 #endif
 
 #ifndef   BUFFER_SIZE_UP
-  #define BUFFER_SIZE_UP                            (1024)  // Size of the buffer for terminal output of target, up to host (Default: 1k)
+  #define BUFFER_SIZE_UP                            (128)  // Size of the buffer for terminal output of target, up to host (Default: 1k)
 #endif
 
 #ifndef   BUFFER_SIZE_DOWN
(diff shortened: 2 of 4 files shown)


nanosonde:
The router object(s) load state checks are missing in bausystembcoupler::configured().
So this needs be fixed by adding them.

Also I am unsure about the AppProgramObject and its role in a coupler. It might be that it is not necessary at all. @thelsing: do you know more about the AppProgramObjects for couplers?

When programming a coupler, does ETS also write something into the AppProgramObject? 
If not, then is clear that „configured()“ will always return false.

However, the suggested change here is not the way to fo IMHO as it needs be fixed properly.

rueckix:
Just to clarify my observation. I don't use KNX Secure, so ETS is not writin into  "_secIfObj". 
Without the USE_DATASECURE flag, the code works (at least in my case).

Not sure what the missing checks are. There are a couple of "loadState() == LS_LOADED" checks. They at least appear to determine whether ETS has programmed the device.

Can you elaborate on what might be missing here?

nanosonde:
Ok, if you do not use KNX Data Secure then you are supposed to comment the line "#define USE_DATA_SECURE" in the config.h as you suggested. 

However, I would like to keep using it as the default. If you do not want to use the defaults from config.h, you could always define NO_KNX_CONFIG and provide all defines via command line switches yourself.
So there is no need to patch change the default in general as suggested here. @thelsing Do you agree or do you also want a new default with Data Secure disabled? It might not be obvious for the "normal" user of the lib to change that a change is required.

BTW: that is also the reason why KNX manufacturer provide two KNX application to program to a KNX Secure capable device: one with data secure support and one without data secure support.

If we use a new default without Data Secure then we have to adjust the CI tests to also compile a variant with Data Secure support.

Concernign the missing checks: you mentioned "bool BauSystemBCoupler::configured()". So this is for a KNX coupler mask only.
A coupler always contains at least one router object depending on the coupler model version that is used. I forgot to add the check if the router object (table) is loaded. Coupler stuff is still experimental. We still have to extend the CreateKnxProd tool a little bit to handle coupler masks and add a few coupler specific XML attributes to the knxprod file for a coupler.

nanosonde:
Another remark/rule:
If the knxprod does NOT indiciate that a KNX device (whether coupler or not) is supporting KNX data secure, then the firmware shall not be compiled with "#define USE_DATA_SECURE".  
USE_DATA_SECURE is NOT an optionally!

knxprod file with data secure -> build firmware with USE_DATA_SECURE
knxprod file without data secure -> build firmware without USE_DATA_SECURE

Mixing is not allowed!

thelsing:
I'm in favor of disabling DATA_SECURE and enable via CMake/platform-io. So it "just works" for arduino-ide users and the more advanced users can tweak the config in platformio.

rueckix:
Apologies, I started committing to the wrong branch. Other changes are mixed in with this PR. These should be rejected to keep things clean.

rueckix:
Will clean up and re-submit.

PR #112, opened 2020-12-06
Fix: "bool BauSystemBCoupler::configured()"

rueckix:
(recreating a clean version of https://github.com/thelsing/knx/pull/108)
Having USE_DATASECURE defined by default breaks installations that do not use KNX Secure.

Reason
Due the the way "bool BauSystemBCoupler::configured()" is implemented, "knx.configured()" will always return "false"

Proposed Solution
Let the user define USE_DATASECURE via compiler flags when required.

### DIFF
--- a/examples/knx-demo-coupler/platformio-ci.ini
+++ b/examples/knx-demo-coupler/platformio-ci.ini
@@ -21,6 +21,7 @@ lib_deps =
 build_flags =
   -DMASK_VERSION=0x2920
   -Wno-unknown-pragmas
+  -DUSE_DATASECURE
 
 ;-----------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


thelsing:
Can you change the platformio-ci.ini files so that data-secure is defined please?

rueckix:
Added to all configs. However, I am not certain if it is meaningful for IP.

nanosonde:
> 
> 
> Added to all configs. However, I am not certain if it is meaningful for IP.

Hi @rueckix 
Data Secure is important for IP as well.
A KNX IP secure device MUST also support Data Secure packets in addition to the the KNX IP secure protocol.
Unfortunately, there is no KNX IP secure implementation at the moment for this stack.

rueckix:
That's what confused me. The flag will not have an effect at the moment. In any case, the flag is now set for all pio targets.

thelsing:
Merged. Thanks.

PR #118, opened 2021-01-07
Fix ambiguous call to attachInterrupt

rueckix:
Recent SAMD arduino core has multiple attachInterrupt overloads. Using the enum type PinStatus instead of a generic uint32_t for the _buttonPinIntteruptOn variable addresses this.


rueckix:
Oh my... Arduino API is undergoing changes and not all platforms are up to date. See discussion at https://github.com/arduino/ArduinoCore-samd/issues/587

thelsing:
Merged. Thanks.

Issue #116, opened 2021-01-03
Incorrect Parameter Encoding (int32)


rueckix:
Trying to configure and read a 32bit int parameter, I am observing that they are not encoded (by ETS 5) as expected and therefore cannot be recovered by this knx stack.

### XML Parameter Type
```
              <ParameterType Id="M-00FA_A-EA61-7B-0000_PT-BlindRuntime" Name="BlindRuntime">
                <TypeNumber SizeInBit="32" Type="unsignedInt" minInclusive="0" maxInclusive="600" />

```
### XML Parameters
```
            <Parameters>
              <Parameter Id="M-00FA_A-EA61-7B-0000_P-1" Name="tBlind1" ParameterType="M-00FA_A-EA61-7B-0000_PT-BlindRuntime" Text="Rolladen 1 Laufzeit [s]" Value="30">
                <Memory CodeSegment="M-00FA_A-EA61-7B-0000_RS-04-00000" Offset="0" BitOffset="0" />
              </Parameter>
              <Parameter Id="M-00FA_A-EA61-7B-0000_P-2" Name="tBlind2" ParameterType="M-00FA_A-EA61-7B-0000_PT-BlindRuntime" Text="Rolladen 2 Laufzeit [s]" Value="30">
                <Memory CodeSegment="M-00FA_A-EA61-7B-0000_RS-04-00000" Offset="4" BitOffset="0" />
              </Parameter>
            </Parameters>

```

### XML Load Procedure

```
<LoadProcedures>
              <LoadProcedure MergeId="2">
                <LdCtrlRelSegment LsmIdx="4" Size="8" Mode="0" Fill="0" AppliesTo="full" />
              </LoadProcedure>
              <LoadProcedure MergeId="4">
                <LdCtrlWriteRelMem ObjIdx="4" Offset="0" Size="8" Verify="true" />
              </LoadProcedure>
</LoadProcedures>
```

### Resulting Encoding
Output of `printHex("", knx.paramData(0), 50);`

1. Value 300 for both parameters
`00 20 01 2C 00 00 01 2C  | 00 0F 13 00 13 00 53 00 4B 07 13 00 13 00 53 00 4B 07 13 00 13 00 5B 00 13 00 13 00 5B 00 5B 00 00 00 FF FF 00 00 FF FF FF FF `

2. Value 10 for both parameters
[...]

PR #119, opened 2021-01-07
Fixing metadata size

rueckix:
The metadata segment in EEPROM/FLASH for storing the application program is not correct. This caused parameters to be written into the metadata segment and vice versa.


rueckix:
#119

thelsing:
Merged Thanks.

thelsing:
Fixes #116

PR #120, opened 2021-01-11
Add: 'Demo example for PZEM004t-V3

zelogik:
Hello,

Working example of reading value from main line (current/voltage/energy/power/frequency), and send it to KNX bus.

Add an automatic progMode, to put device in programmation mode just after the upload of the Samd21

Regards.
ie: Not clear how to simply read the DateTime ....


thelsing:
merged, Thanks.

PR #121, opened 2021-01-18
opt-out global KNX, ability to DIY construct knx object, minor cleanups

defl:
Refactored the knx_facade a bit to make the auto-building of the global knx object and the buttonUp() ISR function opt-in. Also added the ability to quite easily attach another serial (in this case tested with Serial2 of an ESP32).

(Didn't test with opt-in building yet, assumed comments would follow so won't spend time there yet.)


thelsing:
Can you please fix the build checks? Also the default knx instance should be enabled as a default to avoid breaking things for people.

defl:
Fixed tests for platformio locally, will take a look at the failing other ones later.

defl:
All passes now. In the end there is one breaking change for in case where they in Linux have made their own button handler and are calling the now private _toogleProgMode (note the typo in the name which was in the code already). That has been replaced by need to call toggleProgMode(); (b7c67f0)

thelsing:
Merged, thanks.

PR #125, opened 2021-02-13
added alternative GroupObject with less mangagement footprint

mumpf:
Hi Thomas,

this is my improved GroupObject using just 8 bytes management information instead of 19 byte. Can be activated with 
#define SMALL_GROUPOBJECT in config.h
or with compiler option
-DSMALL_GROUPOBJECT

I added also a knx-demo-small-go example project with a commented knx-demo.ino file, which shows how to use this change.

Regards,
Waldemar


thelsing:
What is the motivation of adding another example of knx-demo? I think it should suffice to add some compile test cases to the platform-ci.ini.

mumpf:
No, this would not be sufficient. My change requires a different access to GroupObjects, it would simply not compile with old knx-demo and -DSMALL_GROUPOBJECT.

The new knx-demo is a commented example. It shows the necessary changes. If you think, that all other files, like xml and knxprod files should be removed, this would be ok for me. I just thaugt, the example projects are always self contained, that's why I made a copy of the full knx-demo project.

thelsing:
If only the sketch is different, can you put it in the same directory as the normal knx-demo sketch?

mumpf:
Yes of course, I can do it. I just thaught, that the arduino IDE supports just one ino-File per project and I didn't want to break the demo project.

thelsing:
I have no problem with that. Code/File duplication is worse. I think anybody should be able to cope with two ino files in the directory.  Or you just remove the duplicated files, and add a note that the xml+knxprod of the normal knx-demo project should be used.

mumpf:
Ok, I removed all duplicate files, but I kept the new project directory knx-demo-small-go and added a comment in the ino-file, that knxprod files can be taken from the original knx-demo project. I also kept the modified platformio.ini file as an example how to compile with platformio.

thelsing:
Merged. Thanks.

Issue #90, opened 2020-09-14
Implement way to generate knx-serial number from platform


mjm987:
When more than one KNX-RF device is active, nodes where knx.progMode() is off are unexpectedly programmed as well when programming the physical address of a new node. 
So it seems the state of knx.progMode() has no effect.

Shouldn't  prgMode() checked in following methods of bau27B0.cpp:
void Bau27B0::domainAddressSerialNumberWriteIndication()
void Bau27B0::individualAddressSerialNumberWriteIndication()

(Btw: to handle progMode without a button I tried activating knx.ProgMode(true) in the setup() of *.ino and disable it again in the *.ino loop() after knx.configured() && at least 1 Min passed since reset)

nanosonde:
@mjm987 
Hi! You seem to be the first one who actually tries this. :)
While implementing KNX-RF I have never tested this to be honest.

Do you have multiple DIY KNX-RF devices? If so, did you set a new serial number for each device so that they differ?
All application layer services which have something with "SerialNumber" in their name do NOT take care about the prog button.
Compare https://github.com/thelsing/knx/blob/master/src/knx/bau_systemB.cpp#L447 and https://github.com/thelsing/knx/blob/master/src/knx/bau_systemB.cpp#L441

nanosonde:
BTW: The KNX serial number consists if 4 bytes. So a KNX manufacurer can sell 4294967295 (0xFFFFFFFF) devices.
Ok, 4294967295 - 2 devices. 0xFFFFFFFF and 0x00000000 are not valid. :)

mjm987:
Hi @nanosonde, thx for response!
Yes I'm trying with a couple of nodes to build a garden sprinkler with valves and humidity sensors connected by KNX-RF.
And no, I didn't change the serial number and let it to "Default to KNXA (0xFA)".
According to the referenced comment in the source I expected it is ok for ETS to have always the same serial so I will possible try to generate a random on initial start or first lookup of the serial.
(I didn't like the idea to make a individual program change for every new board.)
To handle the problem temporarily, I hardcoded ownAddress on newly devices and prevented to program the physical address by ETS.

nanosonde:
> And no, I didn't change the serial number and let it to "Default to KNXA (0xFA)".
The manufacturer part must be kept as 0xFA.
Otherwise the DIY generated product databases won‘t work when ETS reads the manufacturer.

[...]

PR #131, opened 2021-04-12
generate unique serial number (#90)

OutOfSync1:
* see https://github.com/ricaun/ArduinoUniqueID

* calculated from ESP.getEfuseMac() on ESP32

* ESP.getChipId() on ESP8266

* SERIAL_NUMBER_WORD_0-3 on SAMD

* HAL_GetUIDw0-2() on STM32

* defaults to 0x01020304 on other platforms

So far only tested on ESP32. But SAMD seems to also work. We should try to add support for the other platforms (linux and CC1310).


thelsing:
Merged. Thanks.

PR #132, opened 2021-04-12
fixes to enable partial programming

OutOfSync1:
(thanks to mumpf and proggerKA at KNX-UF!)

* add PID_MCB_TABLE to TableObject

* add CallBackProperty to send PID_MCB_TABLE

* calculate crc checksum using Crc16Citt when state switches to LOAD_COMPLETED

* add crc to save(), restore() and saveSize() to save crc to flash

* add CallBackProperty for read and write of PID_PROG_VERSION to ApplicationProgramObject

* create overrides for save, restore, and saveSize in ApplicationProgramObject to save _programVersion to flash

Only tested on ESP32. A forked version works on SAMD. 

I saw that there is already a version of CRC16Citt in bits.h. I have not tested if the one in bits.h gives the same results as the new method in table_object.cpp. The new method was tested by proggerKA to calculate the correct checksum as in the example in the KNX spec. This could probably be combined and one of the methods be removed.


nanosonde:
What about the router object?
This needs refactoring too in this case, as it also implements the MCB table.

nanosonde:
The crc16citt in bits.h is used exactly for this purpose: MCB Table CRC.

src/knx/application_program_object.cpp
```diff
@@ -11,7 +11,16 @@ ApplicationProgramObject::ApplicationProgramObject(Memory& memory)
...
+            [](ApplicationProgramObject* io, uint16_t start, uint8_t count, uint8_t* data) -> uint8_t {
+                pushByteArray(io->_programVersion, 5, data);
+                return 1;
+            },
+            [](ApplicationProgramObject* io, uint16_t start, uint8_t count, const uint8_t* data) -> uint8_t 
+            { 
+                memcpy(io->_programVersion, data, 5);
+                return 1;
+            }),
         new CallbackProperty<ApplicationProgramObject>(this, PID_PEI_TYPE, false, PDT_UNSIGNED_CHAR, 1, ReadLv3 | WriteLv0,
```
thelsing, line 24: Is there any reason to not use the Dataproperty?

OutOfSync1: I initially tried with the DataProperty but this did not work and I am not sure why. The value was returned correctly during full programming but not when I did partial programming. ETS always stopped with an error "trying to access protected memory" because the property was not returned (count=0). IIRC when I used the CallbackProperty without saving _programVersion to flash I had the same problem. I will test this again and report.

OutOfSync1:
> 
> 
> The crc16citt in bits.h is used exactly for this purpose: MCB Table CRC.

OK, I will use the method in bits.h then. Thanks!

src/knx/table_object.cpp
```diff
@@ -285,3 +300,19 @@ void TableObject::initializeProperties(size_t propertiesSize, Property** propert
...
+        return 0;
+    if (length == 0)
+        return 0;
+    uint16_t crc = initialValue;
+    for (int i = 0; i <= length; i++)
+    {
+        crc = (uint16_t)((crc << 8) ^ Table[((crc >> 8) ^ (0xff & data[i]))]);
+    }
+    return crc;
+}
```
thelsing, line 318: This should go to its own file with its own header.

thelsing: Ok. Forget that. Just use the one of bits.c

OutOfSync1:
> What about the router object?
> This needs refactoring too in this case, as it also implements the MCB table. The router object inherits from TableObject.
> 
> See
> 
> https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx/router_object.cpp#L87

Yes, I did not see this before, thanks. It seems everything was already there and just needs to be added to other parts, too :)

I think I will move the router object code to table object. Unfortunately I cannot test the router functionality to see if this breaks things.

Do you think the crc needs to be saved in the flash? We had a bit of discussion in the KNX-UF about this. [https://knx-user-forum.de/forum/%C3%B6ffentlicher-bereich/knx-eib-forum/diy-do-it-yourself/1635521-wie-kommuniziert-die-ets-mit-knx-ger%C3%A4ten-z-b-beim-partiellen-programmieren?p=1638277#post1638277](url)

etrinh:
Hi, thank you for the improvements!
Just a remark, not related to this PR but #131. I noticed you used printf function. Do you still need this debug log? If yes, is it possible to use print/println instead?
Having printf can add few kB on footprint, it would be nice to have an option to disable it when not required which is doable with print functions and a define.
Also, the current crc16citt doesn't use LUT, so it is better when room is expensive ;)

thelsing:
I don't think that the crc should be saved. Wouldn't it be fine to don't save it as a member variable but calculate it on the fly when requested (and the current loadstate should have a crc). I think error conditions need to be checked. What happens if the object is in state "Unloaded" and somebody reads the MCP data? Currently you return the old crc and claim to the caller that its valid.

OutOfSync1:
I am wondering what the best way is to get the segment size. It is not save anymore after allocating.
See
 https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx/table_object.cpp#L65-L83

We need to know the size of the data without metadata.

OutOfSync1:
> 
> 
> I don't think that the crc should be saved. Wouldn't it be fine to don't save it as a member variable but calculate it on the fly when requested (and the current loadstate should have a crc). I think error conditions need to be checked. What happens if the object is in state "Unloaded" and somebody reads the MCP data? Currently you return the old crc and claim to the caller that its valid.

yes, that is a problem. A CallbackProperty should be OK then, in fact this was implemented in the first version. It is just specified in the KNX spec that the crc should only be calculated when LOADING->LOADED. That is also the way it currently is implemented in RouterObject. 
https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx/router_object.cpp#L450-L459

I tried the same approach without saving to flash, however after a restart the PID_MCB_TABLE property is not set because updateMcb() has not been called, and partial programming does not work. We need to somehow make sure we can always return a value, even when LOADING->LOADED has not occurred. That's why it might be best to save it to flash. Data (and crc) should not change without reprogramming and new calculation of crc then.

thelsing:
Well you can cheat ;) If somebody wants the MCP and you are in state "Loaded" you calculate the crc and return it with "crcValid". Otherwise don't return it. We don't need it in flash or ram when nobody wants it. I think the spec is from time when it took some time to calculate the crc on 8-bit MCUs. But I don't have a strong preference for this.

OutOfSync1:
I tried to add the suggestions. So far I did not yet touch the RouterObject but changed TableObject. The obstacle here is how to get hold of the size of data(). TableObject::saveSize() is not correct if we want to know how large e.g. the ApplicationObject::saveSize() is. The older version saved the size of the memory block but now this seems impossible. Perhaps I overlooked something, this is all new to me. Do you have suggestions?

RouterObject has the same problems as hinted here in the comment:
https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx/router_object.cpp#L87

OutOfSync1:
I re-implemented saving of the allocation size of data(), this is now also saved and restored in flash. The value is now used for CRC calculation and to return the correct segment size value in the MCB table. The values seen in ETS during programming now make much more sense.

Something similar could be added to RouterObject().

src/knx/application_program_object.h
```diff
@@ -7,9 +7,15 @@ class ApplicationProgramObject : public TableObject
...
+    const uint8_t* restore(const uint8_t* buffer) override;
+    uint16_t saveSize() override;
     uint8_t* data(uint32_t addr);
     uint8_t getByte(uint32_t addr);
     uint16_t getWord(uint32_t addr);
     uint32_t getInt(uint32_t addr);
     double getFloat(uint32_t addr, ParameterFloatEncodings encoding);
+    
+	private:
+    uint8_t _programVersion[5] = { 0, 0, 0, 0, 0};
```
thelsing, line 20: Is there a reason to keep _programVersion as a member variable? We can save 5 Byte if don't need it ;)

src/knx/table_object.cpp
```diff
@@ -267,6 +273,21 @@ void TableObject::initializeProperties(size_t propertiesSize, Property** propert
                     pushInt(obj->tableReference(), data);
                 return 1;
             }),
+        new CallbackProperty<TableObject>(this, PID_MCB_TABLE, false, PDT_GENERIC_08, 1, ReadLv3 | WriteLv0,
+            [](TableObject* obj, uint16_t start, uint8_t count, uint8_t* data) -> uint8_t {
+                if (obj->_state != LS_LOADED)
+                    return 0; // need to check return code for invalid
+                
+                uint32_t segmentSize = obj->_size;
```
thelsing, line 281: I'm a bit confused. Didn't it work with saveSize - foo? Or did you misunderstand my comment? I just wanted to replace "savesize() - 4 -2" with something like "savesize() - sizeof(something) - SOMEDEFINE" so that somebody knows what these numbers mean. You wrote it in the comments, but when code changes, comments often don't.

thelsing (changes requested):

src/knx/application_program_object.h
OutOfSync1: OK, changed in new commit.

src/knx/table_object.cpp
OutOfSync1: Sorry for the confusion. Yes, it did not work before with saveSize() - 4 - 1. I had a closer look at what saveSize() returns and it is InterfaceObject::saveSize() + 5:
https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx/table_object.cpp#L230-L233

InterfaceObject::saveSize() is the sum of saveSize() of all properties with writeEnable():
https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx/interface_object.cpp#L168-L181

For the ApplicationObject this is only PID_PROG_VERSION (which is a DataProperty). Including sizeof(_currentElements) this accounts for 7 bytes (2 + 1*5):
https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx/data_property.cpp#L113-L116

However, I know that I defined the code segment for parameters in the application object as 256 byte (in knxprod):
`<RelativeSegment Id="M-00FA_A-0000-04-0000_RS-04-00000" Name="Parameters" Size="256" LoadStateMachine="4" Offset="0" />`

So far the only way I found to really get the value of 256 for the size of application object data() is to store it after allocTable(). In my earlier tests the partial programming was working, but only because I only programmed 6 bytes of memory parameters out of the maximum possible 256 and this value=6 was lower than the "incorrectly" returned value of 7 that I mentioned above. Now with _size stored it returns the correct values for segment size in the MCB_TABLE during ETS programming.

Sorry for the rather complicated explanation, I hope it is more clear.

thelsing: You should be able to get the correct size. Let me check.

thelsing: Its possible but a little bit if work: You need to add a method to the Memory class for this. See this line: [https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx/memory.cpp#L115]

thelsing: I'm also fine with leaving it as it is. But in this case please add a comment why the _size field is needed. Otherwise I might remove it accidentally a year later. ;)

OutOfSync1: Thanks for checking. I think I understand what you mean. I could add a method "Memory::findBlockSizeOfBlock()" which tries to find a MemoryBlock* of the address and then return the size if its not a nullptr (in this case the address is not in _usedList and it should return 0 indicating that the block is not found).

This method could then be called within TableObject with _memory->findBlockSizeOfBlock(_data).

I will try this, challenge accepted ;)

OutOfSync1: Hmm, in principle this works. However, the alignment is a problem:
https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx/memory.cpp#L155

Here the value of size might be changed. I e.g. got 12 as block->size when allocMemory(10) was initially requested. For CRC calculation this should not matter as data() is allocated up to data()+block->size but the MCB_TABLE segment size would be incorrect then.

thelsing: Ok. Then please add a comment explaining why the _size field is there and why we can't use the size of the memory block and were done.

OutOfSync1: Done. Unfortunately I cannot test what needs to be done to RouterObject. Do we simply have to remove the MCB_TABLE DataProperty and updateMcb() there?

@nanosonde can you comment?

thelsing (approved):

src/knx/table_object.cpp
nanosonde: I will check. Has been a long time since implementation. 😉

nanosonde: Hi!
I have analyzed it. 
Simply removing the MCB_TABLE DataProperty and updateMcb() should be enough.

OutOfSync1: Thanks nanosonde! I removed this as suggested.

thelsing (approved):

thelsing:
Merged. Thanks for the effort.

mumpf:
Hi,

should it somewhere be mentioned, that partial programming needs the correct LoadProcedures in application xml file?

            <LoadProcedures>
              <LoadProcedure MergeId="2">
                <LdCtrlRelSegment LsmIdx="4" Size="0" Mode="1" Fill="0" AppliesTo="full" />
                <LdCtrlRelSegment LsmIdx="4" Size="0" Mode="0" Fill="0" AppliesTo="par" />
              </LoadProcedure>
              <LoadProcedure MergeId="4">
                <LdCtrlWriteRelMem ObjIdx="4" Offset="0" Size="0" Verify="true" AppliesTo="full,par" />
              </LoadProcedure>
              <LoadProcedure MergeId="7">
                  <LdCtrlLoadImageProp ObjIdx="1" PropId="27" />
                  <LdCtrlLoadImageProp ObjIdx="2" PropId="27" />
                  <LdCtrlLoadImageProp ObjIdx="3" PropId="27" />
                  <LdCtrlLoadImageProp ObjIdx="4" PropId="27" />
              </LoadProcedure>
            </LoadProcedures>

Otherwise these changes will have no effect.
Regards, Waldemar

OutOfSync1:
You raise a valid point. We could add it to one of the examples, e.g. the knx-demo. Or it could also be added to the wiki. As far as I see the wiki is still empty and this could make a good start to add more info. What do you think?

etrinh:
Why not updating sister project CreateKnxProd to generate correct xml/knxprod out of the box?
I did at my side, I can make a PR. Also, are all the 4 LdCtrlLoadImageProp required? Does only the one with ObjIdx="4" be enough?

@thelsing what is the status about CreateKnxProd, I saw there is a new devel branch, on which of them should I post PR? Changes seem to be incompatible with each other due to changes in master XML def.

nanosonde:
@etrinh 
Concerning CreateKnxProd devel branch: the most recent schema was added. As there was no ready-to-use converter available, I changed the CreateKnxProd in such a way that it does not rely on the converter DLLs anymore but directly talks to the DLLs which are available with a normal ETS5 installation.

thelsing:
@OutOfSync1  Feel free to do both. I enabled the wiki because of a user request. But nobody wrote anything yet. @mumpf You could also add link/description for you knxprod generation tool there.

thelsing:
@etrinh Because of the changes of the xml scheme I didn't release it yet. I was planing to create a conversion from the old xml. But I guess this won't happen. I will just create a last release of the main branch and merge the devel branch. So: pull requests go to the devel branch.

etrinh:
I just created a PR (https://github.com/thelsing/CreateKnxProd/pull/26) in CreateKnxProd 
Feel free to comment

PR #133, opened 2021-04-14
fix userdata saverestore

OutOfSync1:
restore() in knx_facade.h needs to be implemented using const uint8_t* to override save_restore.h as seen here:
https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx/save_restore.h#L31-L34

Currently the default declaration in save_restore.h is called because knx_facade.h does not override it correctly:
https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx_facade.h#L354-L360
https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx_facade.h#L43
https://github.com/thelsing/knx/blob/e57bbf9dbe25505a9f02d83f111d9df7852c18df/src/knx_facade.h#L274-L277

ESP32 EEPROM library:
For ESP32 the EEPROM library checks if data was modified to avoid unnecessary writes to the flash. This is one with a dirty flag which is set to true in the "write" methods of the EEPROM library or when the initial pointer is requested in ::getDataPtr(). Since we do not use those write methods and we only request the pointer once (and save it internally), the first commit to flash works but subsequent ones don't. Just calling "EEPROM.getDataPtr()" before "EEPROM.commit()" sets the dirty flag to true and forces writing of the flash.

### DIFF
--- a/examples/knx-bme680/knx-bme680.ino
+++ b/examples/knx-bme680/knx-bme680.ino
@@ -25,7 +25,7 @@
 void checkIaqSensorStatus(void);
 void errLeds(void);
 uint8_t* saveBme680State(uint8_t* buffer);
-uint8_t* loadBme680State(uint8_t* buffer);
+const uint8_t* loadBme680State(const uint8_t* buffer);
 void triggerCallback(GroupObject& go);
 void updateState();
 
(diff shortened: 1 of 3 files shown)


thelsing:
Merged. Thanks.

PR #137, opened 2021-05-05
Add new KNX_NO_STRTOx_CONVERSION define for footprint reduction

etrinh:
KNX_NO_STRTOx_CONVERSION  do not perform conversion from KNXValue(const char*) to other types. It mainly avoids the linkage of strtod.
And add some descriptions in config.h


thelsing:
Merged. Thanks.

PR #141, opened 2021-06-04
add KNX IP discovery support 

croghostrider:
#9 added KNX IP discovery support for ESP8266 and ESP32 boards.
Tested with ETS 5.7.4 and an ESP32 Board.


thelsing:
Merged. Thanks.

PR #145, opened 2021-07-08
added support for RP2040 (Raspberry Pi Pico)

SirSydom:
tested to work with Raspberry Pi Pico, demo sektch is running, programming with ETS works and device is sending knx telegrams


thelsing:
Merged. Thanks.

PR #157, opened 2021-12-22
Returning correct value for ULongType.

mptei:
KNXValue::ulongValue() returns wrong value for type ULongType. This PR fixes this.


thelsing:
Merged. Thanks.

PR #158, opened 2021-12-22
Double code removal

mptei:
Somehow in dpt convert code is doubled. This PR removes the doubled code.


thelsing:
Merged. Thanks.

Issue #89, opened 2020-09-11
Error Messages while programming Application with ETS


Space-Farmer:
[LOG_Adress&App_10ms.txt](https://github.com/thelsing/knx/files/5210481/LOG_Adress.App_10ms.txt)
[LOG_Adress&App.txt](https://github.com/thelsing/knx/files/5210482/LOG_Adress.App.txt)

Hello,

my Setup:
IDE: Platform IO Core 5.0.1 Home 3.3.0
MCU: SAMD21 developement Board (Arduino Zero clone without debug circuit, wemo samd21)
Interface: MicroBCU (OnSemi 5120)
SW: ETS 5.7.2
KNX interface: knxd on BBB (ingBudde shield)

Additional Equipment:
Several Isolated Serial to USB-Sticks
Oscilloscope
J-link debugger

Version from GIT: TAG: 1.2.0 Application knx-Demo

Issue: Programming physical Address working normal, but during programming Application severel errors are shown in serial console.

`got unexpected L_DATA_CON<\r><\n>
got U_STATE_IND: 0x17<\r><\n>
got U_STATE_IND: 0x17<\r><\n>
got U_STATE_IND: 0x17<\r><\n>
got U_STATE_IND: 0x17<\r><\n>
got U_STATE_IND: 0x17<\r><\n>
L_DATA_CON not received within expected time<\r><\n>
...`
My current status of investigation:
HW: 
  Pinning fits.
  Signals RX/TX checked. Levels Ok, rise time Very fast, timing perfect to 19,2kHz.
  Voltage supply. No Votage drops on SAMD21 or NCN5120 side.
SW:
  Tried several "Board" Options (adafruit_feather_m0 [here my debug serial is not working],sparkfun_samd21_dev_usb,zeroUSB) 
  Update to the latest "MASTER" leads to not getting PA programmed, on ESP32 Board-> cyclic resets. -> Back to 1.2.0
  removed comments inside ReadUART
  Logged this data
  Additional I logged this data with newline every 10ms to get an idea what is sent within which time. Please find attached these  files

Do you have an idea where to look? 

Please help me.

thelsing:
I'm quite clueless on the electronic part myself. You can try a simple sketch like https://knx-user-forum.de/forum/öffentlicher-bereich/knx-eib-forum/diy-do-it-yourself/33016-arduino-am-knx?p=1141860#post1141860 to see if your wiring is correct.

Space-Farmer:
[...]

Issue #83, opened 2020-08-26
Multiple device in programming mode when my device is alone


ajlajlajl:
Hi
I've got a strange problem with my device. When ETS is looking for device address (when downloading address or in diagnostics's programming mode; if theres another KNX device on TP bus together with my device, theres no problem. But if theres only my device on TP Bus, ETS detects 2 device and fails (on downloading) saying theres more than one device in programming mode, and in Diagnostics also shows two device (*2) with same address.

This is log from Group Monitor for one request:
```
<CommunicationLog xmlns="http://knx.org/xml/telegrams/01">
  <Telegram Timestamp="2020-08-26T13:33:10.6411059Z" Service="L_Data.con" FrameFormat="CommonEmi" RawData="2E00B0E000010000010100" />
  <Telegram Timestamp="2020-08-26T13:33:10.6893658Z" Service="L_Data.ind" FrameFormat="CommonEmi" RawData="2900B0D011020000010140" />
  <Telegram Timestamp="2020-08-26T13:33:10.7272676Z" Service="L_Data.ind" FrameFormat="CommonEmi" RawData="2900B0D011020000010140" />
  <Telegram Timestamp="2020-08-26T13:33:10.7811246Z" Service="L_Data.ind" FrameFormat="CommonEmi" RawData="2900B0D011020000010140" />
</CommunicationLog>
```
As you see each request receives 3 response...

This is debug output of my device (per request):
```
dataBroadcastIndication
dataBroadcastRequest
dataBroadcastIndication
dataBroadcastRequest
dataBroadcastConfirm
dataBroadcastIndication
dataBroadcastRequest
dataBroadcastConfirm
dataBroadcastIndication
dataBroadcastRequest
expected L_DATA_CON not received
dataBroadcastConfirm
got U_CONFIGURE_IND: 0x11
got UNEXPECTED: 0x2
got L_ACKN_IND
got L_ACKN_IND
got UNEXPECTED: 0xE1
got U_CONFIGURE_IND: 0x1
got L_ACKN_IND
got UNEXPECTED: 0xDC
got U_STATE_IND: 0x17
got unexpected L_DATA_CON
```

any idea whats going on here??

Thanks

thelsing:
I wouldn't exclude a bug in the tpuart-code. I will check if I can reproduce this. How do you connect ETS with the bus?

ajlajlajl:
> 
> 
[...]

PR #161, opened 2021-12-28
Use full duplex in tpuart_data_link_layer.

mptei:
tl;dr: PR implements full duplex in tpuart layer. Stay in loop in time critical handling.

The tpuart-layer is implemented in half duplex with a common state for sending and receiving. This works only with very low bus load and leads to many transmission errors. The tpuart communication is full duplex. While uploading a layer-2 packet to the TPUART it may receive packets from the bus. Also it has to acknowledge the received packet as soon as possible while uploading.

Commit 'TPUART full duplex handling' changes the tpuart-layer to be full duplex. For that it uses two state machines. One for sending and one for receiving. It also recognizes end-of-packet (EOP) gaps. The TPUART signals the end of a packet with such a gap. This must be used to stop processing a layer-2 packet receiption and can be used to resynchronize to the stream.

I have (stress) tested this PR with success. As long as the loop is called fast enough the communication is very stable. There are no lost packets.

If the load increases and the loop is called not fast enough it came to many receiption errors. The problem here is that the receiption of a layer-2 packet is very time critical. The EOP is only 2-2.5ms wide. To deal with this, the commit 'Stay in loop when RX_L_DATA.' stays in loop as long as a layer-2 receiption is ongoing. This makes receiption stable on higher load.

I have tested this PR against a TPUART2 and an NCN5120. 

Errors as seen here: https://github.com/thelsing/knx/issues/89#issuecomment-697998765 or https://github.com/thelsing/knx/issues/83#issue-686347861 are solved with this PR.

The pattern:
```
expected L_DATA_CON not received
got U_CONFIGURE_IND: 0x11
got UNEXPECTED: 0x2
got L_ACKN_IND
got L_ACKN_IND
got UNEXPECTED: 0xE1
got U_CONFIGURE_IND: 0x1
got L_ACKN_IND
got UNEXPECTED: 0xDC
got U_STATE_IND: 0x17
got unexpected L_DATA_CON
```
means nothing else, that when the state machine is waiting for L_DATA_CON another layer-2 packet rushes in. The "unexpected L_DATA_CON" is then the message the state machine was waiting for.

Note: In my opinion the tpuart layer must use interrupts to deal with the UART communication. This would make the receiption much more stable and would avoid loop hogging.

### DIFF
--- a/src/knx/tpuart_data_link_layer.cpp
+++ b/src/knx/tpuart_data_link_layer.cpp
@@ -11,6 +11,9 @@
 #include <stdio.h>
 #include <string.h>
 
+// Activate trace output
+//#define DBG_TRACE
+
 // NCN5120
 //#define NCN5120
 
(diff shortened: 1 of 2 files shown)


thelsing:
Merged Thanks.

thelsing:
I'm not sure if you can access the serial interrupts without messing the arduino Serial objects buffers. A better solution would be to use an RTOS and give the knx task a high priority.

PR #162, opened 2021-12-29
fix constant comparison

croghostrider:
Hi thelsing,

maybe you should activate LGTM and codefactor apps for better code quality.


thelsing:
I set up LGTM and codefactor.

PR #167, opened 2022-01-16
Added support for DPT 251.600 RGBW

mptei:
The value is delivered as  uint32_t. The colors are identical to RGB with the white channel added in the most sigificant byte. Resulting in WRGB which is identical to the usage in the neopixel library.

The color mask is get/set via type index 1.

### DIFF
--- a/src/knx/dpt.h
+++ b/src/knx/dpt.h
@@ -357,6 +357,7 @@
 #define DPT_FlaggedScaling Dpt(239, 1)
 #define DPT_CombinedPosition Dpt(240, 800)
 #define DPT_StatusSAB Dpt(241, 800)
+#define DPT_Colour_RGBW Dpt(251, 600)
 
 class Dpt
 {
--- a/src/knx/dptconvert.cpp
+++ b/src/knx/dptconvert.cpp
@@ -126,6 +126,9 @@ int KNX_Decode_Value(uint8_t* payload, size_t payload_length, const Dpt& datatyp
         // DPT 239.* - Flagged Scaling
         if (datatype.mainGroup == 239 && datatype.subGroup == 1 && datatype.index <= 1)
             return busValueToFlaggedScaling(payload, payload_length, datatype, value);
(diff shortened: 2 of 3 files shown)


thelsing:
Merged. Thanks.

dev-git-usr:
Hi,
very funny, i did yesterday a implementation of 251.600 but today there were some merge-conflicts ;)
https://github.com/dev-git-usr/knx/tree/DPT-251_600
Wouldn´t it be better to change the order to RGBW like the Spec of the datatype?
https://www.knx.org/wAssets/docs/downloads/Certification/Interworking-Datapoint-types/03_07_02-Datapoint-Types-v02.02.01-AS.pdf
I think the order WRGB is confusing for everyone using it.

mptei:
If one has used DPT232.600 RGB with B in the least significant byte it seems naturally not to replace B with W and G with B ..., but to add the white channel in the empty msb.

In the Arduino world I found the Neopixel library using the colors in WRGB order. I've seen this as "prior art". But I don't know if there are other popular libraries, probably in other worlds, which has a broader user base.

Looking at the DPT3.007 implementation where the value could be delivered as uint8_t, but is splitted into the step code and the direction. This is a common thing to do when using this type. The implementation of DPT251.600 follows this pattern by returning the RGBW as one value and the 4 mask bits as another value.

What is your use case? And why seems a uint64_t handy?

mptei:
On the other hand I can follow your argumentation that it might be unexpected to get the color bytes reordered. I would agree to deliver the color value as RGBW.

dev-git-usr:
As you said using a uint64_t is not very handy. That´s why i mentioned only the order of Bytes.
I think your implementation with separation of mask bits is much better.
I recommend changing to RGBW because it´s standard and also defined for DPT251.600 but also adding a comment that it´s WRGB is ok for me.

mptei:
Again I agree to you that the average user will expect the byte order to be according the KNX standard and not according to another library. I will create another pull request to change that.

PR #168, opened 2022-01-18
DPT251.600 Changed color byte order from WRGB to RGBW (R MSB).

mptei:
Initial DPT251.600 support delivered the color bytes in WRGB order which seemed useful, but is unexpected.
This commit changes it to the RGBW order according to the KNX data type definition.

### DIFF
--- a/src/knx/dptconvert.cpp
+++ b/src/knx/dptconvert.cpp
@@ -813,8 +813,7 @@ int busValueToRGBW(const uint8_t* payload, size_t payload_length, const Dpt& dat
     switch (datatype.index) {
         case 0: // The RGBW value
             {
-                uint32_t rgbw = (unsigned32FromPayload(payload, 0) >> 8)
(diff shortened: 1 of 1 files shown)


mptei:
@dev-git-usr Perhaps you can check if this is what you expect.

thelsing:
Merged. Thanks.

mptei:
I do it this way:
```
    go_RGBW_RGBW_Value.dataPointType(DPT_Colour_RGBW);

    .
    .
    .
    uint32_t newValue = (uint32_t)go.value();
    uint8_t mask_bits = (uint8_t)go.value(Dpt(251,600,1));
```
But I don't know if this is the recommended way.

thelsing:
There is currently no other way. I think it would be better do define structs/classes for those multipart data types, add them to KnxValue and fill them instead of working with those indexes.

PR #171, opened 2022-02-13
More tpuart fixes

mptei:
I found some more problems in the tpuart implementation.

**Commit "tx throttle"**: This throttles sending of bytes to the tpuart. This fixes 2 problems. First, when there is no load on the main loop, it is still possible to fill the software tx-buffer of the serial. This leads to blocking calls on uart-write and therefore to a long stay in knx::loop. Second, when the software tx-buffer is full and we need to acknowledge a just received L2-packet, our acknowledge is put at the end of the buffer and is therefore delayed by a certainly long time. Our acknowledge may come to late then.
By throttling the transmission, the software tx-buffer still contains not more than 2-3 bytes.

**Commit "New state RX_L_ADDR; loop load adaption"**: In my former PR I've introduced waiting in knx::loop until the whole L2-packet is received. This made communication in my use case (with high loop load and therefore long delays between knx::loop calls) much more stable. But this also increases the time staying in knx::loop a lot. This may influence other parts of the software.
The only time critical part though, is the acknowledgement after receiving the address. This commit introduces a new state-machine state RX_L_ADDR which marks the receiving of an L2 packet until the address is available. Now we can stay in the loop only until the address is received. The rest of the L2 packet can be handled more relaxed in the following knx::loop calls.
Also the EOP detection is relaxed to 8ms (instead of 3ms) when receiving the address. This avoids false detection of EOP when knx::loop is called delayed, but still allows to detect the pause after aborting a transmission.

Also the decision, if the code should stay in knx::loop until the address is received, can be influenced now. The default is, that it doesn't stay in loop (as it was before). This is perfect, if the whole software maintains a high loop frequency.

If the macro **KNX_AUTO_ADAPT** is set, the loop delay is measured. When it is low, knx::loop is left early, but when it is high, we stay in knx::loop until the address is received. This setting is good when the loop load may change from one level to another and stays there for a longer time. On the load change communication may be distorted, but it then stabilizes again. If there are only short loop load changes it may lead to communication errors.

If the macro **KNX_WAIT_FOR_ADDR** is set, it stays in knx::loop until the address is received. This setting can deal with higher loop load, but it normally wastes some time.

**Commit "handle all available bytes"** stays in knx::loop as long as we have data in the software rx-buffer. Here it makes no sense to postpone processing to the next knx::loop call, because processing is fast and we never know if the next loop call happens right in time. So it makes communication more stable to handle all data which is already there. This can't lead to overly long stays in knx::loop, because we can process data much faster than the tpuart can deliver more data. The bus is to slow for this.

### DIFF
--- a/src/knx/tpuart_data_link_layer.cpp
+++ b/src/knx/tpuart_data_link_layer.cpp
@@ -96,6 +96,7 @@ enum {
 #define EOP_TIMEOUT           2   //milli seconds; end of layer-2 packet gap
 #define CONFIRM_TIMEOUT       500  //milli seconds
 #define RESET_TIMEOUT         100 //milli seconds
+#define TX_THROTTLE_TIME      1 //milli seconds
 
 void TpUartDataLinkLayer::loop()
 {
@@ -295,6 +296,7 @@ void TpUartDataLinkLayer::loop()
                         }
 
                         // Hint: We can send directly here, this doesn't disturb other transmissions
+                        // We don't have to update _lastByteTxTime because after ACK the timing is not so tight
                         _platform.writeUart(c);
                     }
                 }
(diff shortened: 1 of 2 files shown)


mptei:
Commit "printHex takes to long" removes the call to printHex in error case. I've just seen that this call takes very long.

thelsing:
Merged. Thanks.

PR #173, opened 2022-02-19
Add callback functions for prog mode and improve documentation

mgeramb:
- Add callback functions for indicating prog mode
- Allow to set LED_BUILTIN to another value
- Improved documentation

@thelsing Thank you for providing this library!


thelsing:
Merged. Thanks.

Issue #176, opened 2022-02-20
Support direct reading and writing of flash memory


SirSydom:
the stack shall support the usage of flash memory without any ram buffering to optimze ram usage.
The flash block buffering and writing is done within the stack, the hardware access in the specific platforms.

For legacy support of the EEPROM-Emulation or to use a RAM-Buffered flash approach the EEPROM api to the platforms should stay functional.

SirSydom:
closed with #177

PR #177, opened 2022-02-20
Branch flash pr with large eeprom

SirSydom:
solves #176 

- flash reading/writing logic completed based on the "flash" branch
- added support for plattforms using only EEPROM-Emulation API
- added direct flash support for RP2040 plattform
- added ram-bufferd flash support to RP2040 plattform


src/knx/platform.h
```diff
@@ -49,21 +54,60 @@ class Platform
...
+
+    //Memory
+
+    // --- Legacy support only. Do not use for new plattforms ---
+    // --- can be remove if all plattforms have been changed to support the NonVolatileMemory functions
+    virtual uint8_t* getEepromBuffer(uint16_t size);
+    virtual void commitToEeprom();
+      // -------------------------------------------------------------------------------------------------------
+
+    virtual uint8_t* getNonVolatileMemoryStart();
```
thelsing, line 66: Maybe we could use __attribute__ ((deprecated)) or something like that?

src/rp2040_arduino_platform.cpp
```diff
@@ -91,6 +155,73 @@ void RP2040ArduinoPlatform::commitToEeprom()
...
+
+    flash_range_program(KNX_FLASH_OFFSET + pageNumber * flashPageSize(), data, flashPageSize());
+
+    rp2040.resumeOtherCore();
+    interrupts();
+}
+
+void RP2040ArduinoPlatform::writeBufferedEraseBlock()
+{
+    if(_bufferedEraseblockNumber > -1 && _bufferedEraseblockDirty)
```
thelsing, line 210: I think a comment that this is only overwritten because to optimize stuff would be helpfull. Otherwise new platforms will needlessly copy this.

src/rp2040_arduino_platform.h
```diff
@@ -13,11 +23,36 @@ class RP2040ArduinoPlatform : public ArduinoPlatform
     void setupUart();
 
     // unique serial number
-    uint32_t uniqueSerialNumber() override;
+    uint32_t uniqueSerialNumber(); //override;
```
thelsing, line 26: Why was the override removed?

SirSydom: when i started with the flash branch master was not merged into flash branch and uniqueSerialNumber not defined (so the override caused a compile error).
override is optional anyway, so i removed it. It can be added again, no problem.

src/rp2040_arduino_platform.cpp
SirSydom: the comment is in src/rp2040_arduino_platform.h

src/knx/platform.h
SirSydom: as i wrote that comment I did not realize that the EEPROM Emulation API is also usefull for implementing a plattform with a RAM-Buffered Flash.
I added the option to use ram buffered flash in the rp2040 plattform by using the EEPROM-API.
So I would now remove change the comment to respect that

src/knx/memory.cpp
```diff
@@ -84,29 +91,37 @@ void Memory::readMemory()
 
 void Memory::writeMemory()
 {
-    uint8_t* buffer = _data;
-    buffer = pushWord(_deviceObject.manufacturerId(), buffer);
-    buffer = pushByteArray(_deviceObject.hardwareType(), LEN_HARDWARE_TYPE, buffer);
-    buffer = pushWord(_deviceObject.version(), buffer);
+    // first get the necessary size of the writeBuffer
+    size_t writeBufferSize = 16;
```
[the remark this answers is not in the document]
SirSydom: I have no idea.
The change was made by you before I started to work on the flash branch ;)
https://github.com/thelsing/knx/commit/4b873aebd60ec14521826453e2c3ba74726fcdb9

src/knx/platform.h
thelsing: Ok.

src/rp2040_arduino_platform.h
thelsing: Ok. Please do so.

src/rp2040_arduino_platform.cpp
thelsing: Sorry I missed that.

src/knx/memory.cpp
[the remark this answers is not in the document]
thelsing: I think it should be alignToPageSize(_metadataSize) as in readMemory.  Not sure how 16 came to be.

src/knx/platform.h
SirSydom: done

src/rp2040_arduino_platform.h
SirSydom: done

src/knx/memory.cpp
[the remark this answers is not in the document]
SirSydom: changed to writeBufferSize = _metadataSize.
No need for alignToPageSize here because the writebuffer must only hold the real data, not the "spacing" data.

thelsing:
Merged. Thanks.

Issue #178, opened 2022-02-21
problems with APDU > 56


SirSydom:
no respond to MemoryWrite on RP2040 plattform when APDU > 56

similar issues exists with SAMD.

Must be investigated.

ExtMemoryWrite with 128Byte worked RP2040 plattform with Core arduino-pico 1.9.1

SirSydom:
a MemoryWrite with 60byte payload is received within the stack.
if verifymode == false, the response is good.
if verifymode == true, the response also has 60byte payload and is not received.

It seems the MemoryResponse is added correctly to the tx queue in the stack.
but when sending it something goes wrong.
that is beyond my knowledge.

@mptei maybe you have an idea as tpuart specialist?

[trace.zip](https://github.com/thelsing/knx/files/8112104/trace.zip)

SirSydom:
from the Trace:
> F604E7C118F1011A0A16A23FA34EA4A5BCA68A7A80A9A0AAA1ABACA2ADA3AEA4AFB0A5B1A6B2A7B3B4A8B5A9B6AAB7B8ABB9ACBAADBBBCAEBDAFBEBFB0C0B1C1B2C2C3B3C4B4C5B5C6C7B6C8B7C9B8CACBB9CCBACDBBCECFBCD0BDD1BED2D3BFD4C0D5D6C1D7C2D8C3D9DAC4DBC55DTX_WAIT_ECHO
C6C7C8C9CACBCCCDCECFD0D1D2D3D4D5D6D7D8D9DADBBDRX_WAIT_START
17got U_STATE_IND: PE


from the datasheet:
pe (protocol error): ‘1’ if an incorrect sequence of commands sent by the host controller is detected

mptei:
I guess you are using an NCN5120? TPUART and TPUART2 have only 64 byte buffers for sending frames. The 60 byte transmission exceeds that limit.

My test system currently has a TPUART2. Therefore it doesn't support longer packets. I'll switch to an NCN5120, but this might take a while.

Yes, that is an protocol error. The trace output is not sufficient to say what's wrong. The bytes send to TPUART are not fully shown (only the L2 bytes, but not the protocol stuff).

But by looking at the code I think this patch may help:
[send_more_than_64.txt](https://github.com/thelsing/knx/files/8117482/send_more_than_64.txt)

[...]

PR #179, opened 2022-02-22
solves #178 bugfix when sending frames > 63 byte, missing masking of …

SirSydom:
…upper bits after sending a U_DataOffset

### DIFF
--- a/src/knx/tpuart_data_link_layer.cpp
+++ b/src/knx/tpuart_data_link_layer.cpp
@@ -594,7 +594,7 @@ bool TpUartDataLinkLayer::sendSingleFrameByte()
         }
 
         if (_TxByteCnt != _sendBufferLength - 1)
-            cmd[0] = U_L_DATA_START_CONT_REQ | _TxByteCnt;
+            cmd[0] = U_L_DATA_START_CONT_REQ | (_TxByteCnt % 64);
         else
             cmd[0] = U_L_DATA_END_REQ | _TxByteCnt;
 
--- a/src/knx/tpuart_data_link_layer.cpp
+++ b/src/knx/tpuart_data_link_layer.cpp
@@ -596,7 +596,7 @@ bool TpUartDataLinkLayer::sendSingleFrameByte()
         if (_TxByteCnt != _sendBufferLength - 1)
             cmd[0] = U_L_DATA_START_CONT_REQ | (_TxByteCnt % 64);
         else
-            cmd[0] = U_L_DATA_END_REQ | _TxByteCnt;
+            cmd[0] = U_L_DATA_END_REQ | (_TxByteCnt % 64);
 
         cmd[1] = _sendBuffer[_TxByteCnt];
 #ifdef DBG_TRACE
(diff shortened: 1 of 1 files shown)


mptei (changes requested):
Why the modulo operation?
Using a simple "& 0x3f" is sufficient here, or?

This change works only up to 127 bytes.  U_L_DATA_END_REQ needs that mask to.

SirSydom:
> Why the modulo operation?
> Using a simple "& 0x3f" is sufficient here, or?

sufficient yes. I doubt it will make any difference in opcode, since compiler optimize such stuff anyway.
But in my view is much more readable.

the function starts with
`uint8_t idx = _TxByteCnt / 64;`
could also be written as `uint8_t idx = _TxByteCnt >> 5;`
but isn't, because it is not very readable.

here _TxByteCnt is split and the lower part is used - _TxByteCnt % 64 gets the upper part.
For me that is absolutely logical.


> U_L_DATA_END_REQ needs that mask to.
you're right. missed that one.

mptei:
> But in my view is much more readable.

This is just the problem. Here we are in bit banging world. Masking via & is easy to understand. Replacing this by % is possible in this specific case, but generally not.

You pointed exactly at the other place what I would change, Shifting bits is easy to understand when a shift operator is used. Expressing this by a division simply hides what's going on here.

We want to shift bits around, mask them and combine them. We don't want to solve mathematical problems.

SirSydom:
>We want to shift bits around, mask them and combine them. We don't want to solve mathematical problems.

bitbanging IS a way of solving mathemtical problems in a very hardware-dependent manner.
using higher level operations is a more abstract approach.

I oriented myself on the used division and I generally use division and modulo for splitting numbers up into smaller numbers with different place values.

But in the end I don't really care how it is solved.
If you make a PR changing the solution to one using pure bit operations I'm really fine.
@thelsing Thomas should decide what fits into his stack.

thelsing:
```
$ cat test_mod.c test_and.c
#include <stdio.h>
int main (int argc, char *argv[])
{
printf("%d d\n", argc %64);
return 0;
}
#include <stdio.h>
int main (int argc, char *argv[])
{
printf("%d d\n", argc & 0x3f);
return 0;
}
$ gcc -O2 -s test_and.c
$ gcc -O2 -s test_mod.c
$ diff -u test_and.c test_mod.c
--- test_and.c  2022-02-23 20:29:07.829008200 +0100
+++ test_mod.c  2022-02-23 20:26:44.699008200 +0100
@@ -1,6 +1,6 @@
 #include <stdio.h>
 int main (int argc, char *argv[])
 {
-printf("%d d\n", argc & 0x3f);
+printf("%d d\n", argc %64);
 return 0;
 }
$ diff -u test_and.s test_mod.s
--- test_and.s  2022-02-23 20:26:16.859008200 +0100
+++ test_mod.s  2022-02-23 20:26:48.939008200 +0100
@@ -10,12 +10,16 @@
 main:
 .LFB11:
        .cfi_startproc
-       movl    %edi, %esi
+       movl    %edi, %eax
        subq    $8, %rsp
        .cfi_def_cfa_offset 16
+       sarl    $31, %eax
+       shrl    $26, %eax
+       leal    (%rdi,%rax), %esi
        leaq    .LC0(%rip), %rdi
-       xorl    %eax, %eax
        andl    $63, %esi
+       subl    %eax, %esi
+       xorl    %eax, %eax
        call    printf@PLT
        xorl    %eax, %eax
        addq    $8, %rsp
```
I also like & and << better in such case. The example also shows thats it's faster. It would be nice if you could replace the /  64 with a shift.

thelsing:
Merged. Thanks.

PR #182, opened 2022-02-24
Using defines in ACK handling

mptei:
The commands and bits in ACK handling were hard coded. Using defines makes it more readable.


thelsing:
Merged. Thanks.

PR #184, opened 2022-02-26
Handling buffer overrun

mptei:
This PR fixes a problem when the main loop gets stuck. In this moment, if TPUART sends us more data, this fills the serial input buffer.
If this data gets to much we can't process it on the next loop call. This is because time to report back U_ACK_REQ is already over.
Therefore it makes no sense to deal with this old data.
Instead the input buffer is flushed and a new packet start is detected via end of packet gap.


thelsing:
Merged. Thanks.

PR #186, opened 2022-02-26
Reimplement SAMD flash storage based on new memory handling

mumpf:
Provides folllowing features
- Get rid of EEPROM emulation
- Use SAMD-Flash without copying parameters to RAM
Allows bigger ETS-Applications on the SAMD platform

The new flash handling with 1024 Byte flash size is now the default. Size of flash can be defined with KNX_FLASH_SIZE and must be mutliple of 1024. New implementation is independent of previously used FlashStorage library.

Old behaviour can be reastablished with USE_SAMD_EEPROM_EMULATION.

### DIFF
--- a/src/samd_platform.cpp
+++ b/src/samd_platform.cpp
@@ -4,17 +4,29 @@
 #include <knx/bits.h>
 
 #include <Arduino.h>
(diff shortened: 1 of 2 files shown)


thelsing:
Merged. Thanks.

PR #187, opened 2022-02-27
Feature: Uninitialized GO handling

mumpf:
After startup, some knx devices might have uninitialized GroupObjects, because there exists no sensible value for such a GroupObject. This is often the case for logic modules (where the output of a logic might depend on multiple inputs, which did not receive a telegram yet) oder for gateway modules, where the output depends on a value from an other bus system, which did not send a value itself.

A GroupValueRead on sucht an uninitialized GroupObject returned the DPT specific initial value, which might cause wrong actions on the KNX bus. This change prevents GroupValueRead requests for uninitialized GroupObjects. As soon as they get initialized through an external value (GroupValueWrite) or an internal value (GroupObject::value(x)), the behaviour is as before.

This change should have no sideeffects, except someone relies on uninitialized (wrong) values after startup AND reads them explicitly with a GroupValueRead. In such a case, this GroupObject should be initialized with the according (wrong) value by the firmware.

### DIFF
--- a/src/knx/group_object.cpp
+++ b/src/knx/group_object.cpp
@@ -74,6 +74,10 @@ bool GroupObject::readEnable()
     if (!_table)
         return false;
 
+    // we forbid reading of new (uninitialized) go
+    if (_commFlag == Uninitialized)
+        return false;
+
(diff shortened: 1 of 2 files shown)


thelsing:
Merged. Thanks.

PR #189, opened 2022-02-27
Feature: additional callbacks for application

mumpf:
This change adds 2 callbacks:
- beforeRestartCallback, which is called immediately before the module is restarted programmatically (i.e. ResetDevice from ETS or restart after programming). It allows to save application specific data before a restart happens.
- beforeTablesUnloadCallback, which is called directly before the ETS programs the module. It allows to save current GroupObject values, before they get lost as soon als Tables are unloded. 
Usually, a beforeRestartCallback is called after beforeTablesUnload, because ETS restarts a module after a programming procedure.

This change has no sideeffects as long as no callbacks are added by the firmware using this stack.

### DIFF
--- a/src/knx/bau.h
+++ b/src/knx/bau.h
@@ -3,7 +3,7 @@
 #include "knx_types.h"
 #include "interface_object.h"
 
-typedef void (*beforeRestartCallback)(void);
+typedef void (*BeforeRestartCallback)(void);
 
 class BusAccessUnit
 {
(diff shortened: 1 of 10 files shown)


mumpf:
To be honest, I have no clue, how to solve the CodeFactor issue?

src/knx/table_object.h
```diff
@@ -3,6 +3,9 @@
 #include "interface_object.h"
 
 class Memory;
+
+typedef void (*beforeTablesUnloadCallback)();
```
thelsing, line 7: Same as above: please start the type with a capital letter.

thelsing (changes requested):

thelsing:
I forgot: please remove the prefixes from the simple properties (get/set). See DeviceObject for examples.

src/knx/bau.h
```diff
@@ -3,6 +3,8 @@
 #include "knx_types.h"
 #include "interface_object.h"
 
+typedef void (*beforeRestartCallback)(void);
```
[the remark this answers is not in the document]
mumpf: Done.

src/knx/table_object.h
mumpf: Done

mumpf:
Also removed prefixes. Thanks.

thelsing:
Merged. Thanks.

PR #194, opened 2022-04-01
GroupObjects can communicate internally

mumpf:
- GroupObjects use callback infrastructure also during write to bus
- external and internal write of a GO can be handled identically
- needs #define INTERNAL_GROUPOBJECT

Without INTERNAL_GROUPOBJECT the behaviour is not changed. Otherwise the callback, which indicates a GO change is called also as soon as an GO value is sent to the KNX-Bus. In most cases such a behaviour is not necessary. I need it for my generic logicmodule to allow generic connections of Output-GO with Input-GO without sending anything to the KNX bus.

The change is as minimal invasive as I could do...

Regards,
Waldemar

### DIFF
--- a/src/knx/bau_systemB_device.cpp
+++ b/src/knx/bau_systemB_device.cpp
@@ -69,7 +69,6 @@ void BauSystemBDevice::sendNextGroupTelegram()
         if (flag != ReadRequest && flag != WriteRequest)
             continue;
 
-#ifdef INTERNAL_GROUPOBJECT
         if (flag == WriteRequest) 
         {
 #ifdef SMALL_GROUPOBJECT
@@ -80,7 +79,6 @@ void BauSystemBDevice::sendNextGroupTelegram()
                 handler(go);
 #endif
         }
-#endif
         if (!go.communicationEnable())
         {
             go.commFlag(Ok);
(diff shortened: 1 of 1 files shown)


thelsing:
Can't you just move the whole callback handling to the front of the function? Without any ifdefs. The code is simpler I don't see any case where this would hurt.

mumpf:
Just to mention possible side effects:
#ifdef SMALL_GROUPOBJECT is necessary for all users, which use the former large Groupobjects. But I think, you ment
#ifdef INTERNAL_GROUPOBJECT.
If I remove this, there is one marginal sideeffect for existing firmwares:
For pure outputs (GO with CRT-Flags - German: KLÜ-Flags), now an additional callback will be raised. This might be a problem for existing firmwares.
Otherwise: The same would also happen without my change, if in ETS someone sets the W-Flag (German: S-Flag) to this GO, so the event handling should already be aware of such a case. Because users do not very often change flags, developers might not be aware of this case.

I can remove #ifdef INTERNAL_GROUPOBJECT, but I just want doblecheck with you if it is ok with this potential sideeffect.

Regards, Waldemar

thelsing:
Yes I'm fine with that. Simplicity rules!

mumpf:
Hi, I removed the #ifdef INTERNAL_GROUPOBJECT parts.
As already explained, SMALL_GROUPOBJECT cannot be removed.

Regards,
Waldemar

PR #198, opened 2022-04-13
Fix #191 by initializing the platform pointer in all KnxFacade constructors

### LINKED ISSUE #191: Dereference of uninitialized pointer in KnxFacade constructor
When using the KnxFacade constructor, with custom initialized bau object (because you want eg a different hardware serial port), the constructor tries to dereference the _platformPtr (via platform()) which is never initialized, crashing the microcontroller. In src/knx_facade.h;

```
    KnxFacade(B& bau) : _bau(bau)
    {
        manufacturerId(0xfa);
        bauNumber(platform().uniqueSerialNumber());
        _bau.addSaveRestore(this);
    }
```
I don't know what the best way of fixing it is since the platform is a protected member of the bau object, so the facade actually cannot access without a getter being added. Alternatively you could pass the platform object via the constructor, but then you risk a different platform object getting passed in than is actually used within bau.

AODtorusan:
This fixes a null pointer de-reference found in issue #191, where one of the constructors in KnxFacade was accessing the platform pointer without it being initialized, crashing the micro-controller.

Its fixed by also explicitly passing in the platform pointer to the constructor so it can be properly initialized and the KnxFacade::platform() method now always works as expected. An alternative would be to assign a static bau number in the constructor but that would leave the platform pointer uninitialized.

### DIFF
--- a/examples/knx-demo-diy/knx-demo-diy.ino
+++ b/examples/knx-demo-diy/knx-demo-diy.ino
@@ -18,7 +18,7 @@
 // the ESP32's secondary UART and late-bind the ISR function in setup().
 Esp32Platform knxPlatform(&Serial2);
 Bau07B0 knxBau(knxPlatform);
-KnxFacade<Esp32Platform, Bau07B0> knx(knxBau);
+KnxFacade<Esp32Platform, Bau07B0> knx(&knxPlatform, knxBau);
 
 ICACHE_RAM_ATTR void myButtonPressed()
 {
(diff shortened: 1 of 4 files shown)


thelsing:
How about adding an accessor to BauSystemB::platform and using this to initialize the platform instead?

AODtorusan:
Thats also fine, see the updated PR code. 

Note that the cast is required since BauSystemB is not templated with the specific Platform implementation. So technically the user can use the wrong type argument for KnxFacade.

thelsing:
Merged. Thanks.

PR #199, opened 2022-04-18
Removed redundant declaration

croghostrider:
Hi thelsing,

it has no effect on the firmware, the compiled firmware is identical.

### DIFF
--- a/src/cc1310_platform.h
+++ b/src/cc1310_platform.h
@@ -17,11 +17,11 @@ class CC1310Platform : public Platform
     void init();
 
     // basic stuff
-    virtual void restart() final;
-    virtual void fatalError() final;
+    void restart() final;
+    void fatalError() final;
 
-    virtual uint8_t* getEepromBuffer(uint16_t size) final;
-    virtual void commitToEeprom() final;
+    uint8_t* getEepromBuffer(uint16_t size) final;
+    void commitToEeprom() final;
 };
 
 #endif //DeviceFamily_CC13X0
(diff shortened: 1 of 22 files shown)


mumpf:
Hi @croghostrider, 

why do you think this declaration is redundant? It allows to inherit and override according methods. Why should this be removed if inheritence is intended?

Regards,
Waldemar

SirSydom:
I did not check for every member if it is the case, BUT, in general, in the derived class the functions does not have to have the virtual specifier, only the base class must.

mumpf:
I am not an expert in C++, so I am happy to learn something. My current understanding is:

Class A with method "virtual x".
Class B:A with method "x override" works fine, this is the comment of @SirSydom 
Class C:B with method "x override" is not possible, because B::x is not virtual. If B would have a method "virtual x override", C would work. So I am talking about multi level inheritence.

If I am wrong, I am fine with this PR.
But if I am right, this PR would remove the ability to derive from already derived classes. Such a thing might be intended, but is - as far as I can see - not the case here.

Regards,
Waldemar

nanosonde:
The keywords „override“ and „final“ implicitly assume a virtual method.
So removing it is ok.

From my personal experience it depends on the project‘s coding style guidelines if the redundant „virtual“ has to be used or not.

croghostrider:
Hi, I have opened this pull request for fixing the [codefactor ](https://www.codefactor.io/repository/github/thelsing/knx/issues?category=Style&groupId=842)styling issues. 

@SirSydom  l think for better code understand will be better a comment.

For me, it will be fine to leave the redundant override/virtual annotation, but then it should be deactivated this check on codefactor.

thelsing:
Learned something new. I didn't know that virtual is already implied by override and final, but it makes sense. Thanks for submitting the PR.

PR #200, opened 2022-04-21
Get rid of occsional fails during ETS programming

mumpf:
- happens just with large applications
- occured mainly on SAMD platform
- new crc16ccitt algorithm without additional buffer
- reorder groupObjTable in memory for less fragmentation

Hi @thelsing,

this PR solves major issues on programming large applicaitons on an SAMD which are prepared for partial programming. At the end of such a programming session ETS is reading property 27 for each programmed table object, which is a CRC16citt of this table. Previous algorithm allocated an additional memory block in RAM for CRC calculation. 
Large applications (like my logicmodule) have a parameter block > 6k, the contained memcpy was overwriting other allocated memory.

This PR reads directly from provided buffer (i.e. flash) for CRC calculation, so it uses less RAM resources. Tests with both algroithms (old and new) returned identical resultss for CRC.

The correction in bau_systemB_device.cpp changes the order of SaveRestores to the order how ETS is writing the tables. This is just for easier debugging of partial programming and might(?) reduce memory fragmentation in memory::_usedList/memory::_freeList. It has no side effects.

Regards,
Waldemar

### DIFF
--- a/src/knx/bits.cpp
+++ b/src/knx/bits.cpp
@@ -124,21 +124,19 @@ uint64_t sixBytesToUInt64(uint8_t* data)
 
 uint16_t crc16Ccitt(uint8_t* input, uint16_t length)
 {
-        uint32_t polynom = 0x1021;
-        uint8_t padded[length+2];
+    uint32_t polynom = 0x1021;
 
-        memcpy(padded, input, length);
-        memset(padded+length, 0x00, 2);
-
-        uint32_t result = 0xffff;
-        for (uint32_t i = 0; i < 8 * (uint32_t)sizeof(padded); i++) {
-            result <<= 1;
-            uint32_t nextBit = (padded[i / 8] >> (7 - (i % 8))) & 0x1;
-            result |= nextBit;
-            if ((result & 0x10000) != 0)
-                result ^= polynom;
-        }
-        return result & 0xffff;
+    uint32_t result = 0xffff;
+    for (uint32_t i = 0; i < 8 * ((uint32_t)length + 2); i++)
+    {
+        result <<= 1;
+        uint32_t nextBit;
+        nextBit = ((i / 8) < length) ? ((input[i / 8] >> (7 - (i % 8))) & 0x1) : 0;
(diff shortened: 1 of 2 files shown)


src/knx/bits.cpp
```diff
@@ -124,21 +124,19 @@ uint64_t sixBytesToUInt64(uint8_t* data)
...
-            uint32_t nextBit = (padded[i / 8] >> (7 - (i % 8))) & 0x1;
-            result |= nextBit;
-            if ((result & 0x10000) != 0)
-                result ^= polynom;
-        }
-        return result & 0xffff;
+    uint32_t polynom = 0x1021;
+
+    uint32_t result = 0xffff;
+    for (uint32_t i = 0; i < 8 * ((uint32_t)length + 2); i++)
```
thelsing, line 130: You process bytes behind the length supplied as a parameter. I don't think that this should be done. I think the caller should provide a padded buffer and add the +2 to length if needed on the caller side.

src/knx/bits.cpp
```diff
@@ -124,21 +124,19 @@ uint64_t sixBytesToUInt64(uint8_t* data)
...
+    {
+        result <<= 1;
+        uint32_t nextBit;
+        nextBit = ((i / 8) < length) ? ((input[i / 8] >> (7 - (i % 8))) & 0x1) : 0;
```
mumpf, line 134: In the original algorithm there were just 2 bytes 0x00 added to the padded field.
The above line just adds the 2 0x00 during access to the array. So the access to the array is just within its size, any access outside of size gives a 0x00.
The main point im my PR is to save RAM. The original data (input) is usually in Flash. If you insist on an external padded buffer, it would require to copy all data to RAM (like in the old algorithm). On a SAMD with a large application there is not enough space in RAM to do so.

src/knx/bits.cpp
mumpf: No, I don't - see my comment on Line 134.

thelsing:
Sorry I missed that. Merged.

PR #205, opened 2022-04-29
Fix stm32 eeprom handling

rueckix:
Addresses a side effect of https://github.com/thelsing/knx/pull/177, which causes the eeprom buffer to be overwritten every time getEepromBuffer is called. This prevented programming of STM32 boards as their initial bus address was lost before programming.


thelsing:
Merged. Thanks.

PR #207, opened 2022-05-27
Fix prog button handling

rueckix:
Addressing https://github.com/thelsing/knx/discussions/206

A button press is now detected as _valid_ if it is longer than `PROG_BTN_PRESS_MIN_MILLIS` and shorter than `PROG_BTN_PRESS_MAX_MILLIS`. Each with reasonable defaults of 50ms and 500ms.

This removes the need for setting edge detection to RISING of FALLING. The interrupt will simply trigger on CHANGE.

### DIFF
--- a/src/knx_facade.h
+++ b/src/knx_facade.h
@@ -406,7 +406,6 @@ template <class P, class B> class KnxFacade : private SaveRestore
     ProgLedOffCallback _progLedOffCallback = 0;
     uint32_t _ledPinActiveOn = LOW;
     uint32_t _ledPin = LED_BUILTIN;
-    uint32_t _buttonPinInterruptOn = CHANGE;
     uint32_t _buttonPin = 0;
     SaveCallback _saveCallback = 0;
     RestoreCallback _restoreCallback = 0;
--- a/src/knx_facade.h
+++ b/src/knx_facade.h
@@ -177,24 +177,7 @@ template <class P, class B> class KnxFacade : private SaveRestore
         _progLedOnCallback = progLedOnCallback;
     }
 
-    /**
-     * returns RISING if interrupt is created in a rising signal, FALLING otherwise
-     */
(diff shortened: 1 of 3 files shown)


src/knx_facade.h
```diff
@@ -423,7 +406,7 @@ template <class P, class B> class KnxFacade : private SaveRestore
     ProgLedOffCallback _progLedOffCallback = 0;
     uint32_t _ledPinActiveOn = LOW;
     uint32_t _ledPin = LED_BUILTIN;
-    uint32_t _buttonPinInterruptOn = RISING;
+    uint32_t _buttonPinInterruptOn = CHANGE;
```
thelsing, line 409: Just get rid of that field and use CHANGE directly. Or is there a need to keep it?

src/stm32_platform.cpp
```diff
@@ -32,16 +32,21 @@ void Stm32Platform::restart()
 
 uint8_t * Stm32Platform::getEepromBuffer(uint16_t size)
 {
-    if (size > E2END + 1)
```
thelsing, line 35: Strictly speaking this change should be in a different PR. But since it's only a small change its ok.

thelsing (approved):

src/stm32_platform.cpp
rueckix: That's not a change. It must have happened as as re-sync a branch with thelsing:master.  Sorry for the slight mess.

The code is identical. 

master: https://github.com/thelsing/knx/blob/master/src/stm32_platform.cpp
patch: https://github.com/rueckix/knx/blob/patch-progbutton/src/stm32_platform.cpp

src/knx_facade.h
rueckix: No need to keep it. I'll kill it.

thelsing:
Merged. Thanks.

Issue #181, opened 2022-02-23
Porting to H8I8O board


pavkriz:
Hello,
did anyone try this library on [H8I8O board](https://www.aliexpress.com/item/1005003124183354.html)?

It's a cheap KNX device having reasonable number of inputs and outputs. 
It has GD32F103CBT6 MCU (STM32 compatible) and TP-UART 2 ver. 5.3 on board. 
So it should be feasible to run this library on this board.

thelsing:
Is there any way to modify the firmware for this board? Is so it could work.

SirSydom:
looks like it is not meant to change the fw since:

"Bieten eine e-mail adresse zu erhalten die datenbank nach der bestellung"

If there is a SWD/SWCLK interface on the board one could change the fw.
the 4 pin D C G V could be such an interface.

pavkriz:
I confirm the 4 pins D C G V are SWDIO, SWDCLK, GND and +3V. Unfortunately it looks SWD is disabled in the running firmware since my ST-LINK is unable to connect to target MCU.

I am not sure if it's feasible to perform Chip Erase without having to access the NRST pin or BOOT0 (which is inconvenient since these are probably not routed to any header).

BTW by "Bieten eine e-mail adresse zu erhalten die datenbank nach der bestellung" they mean they send you an ETS project (from which you can copy the H8I8O device to another project) and a "User manual" PDF in chinese.

SirSydom:
could you upload the project and the manual? would be interesting.

Regarding programming ESP I have no idea...

pavkriz:
I got SWD working with J-Link: https://gist.github.com/pavkriz/aca0851f38bc79789e878c7aa3777ece
ST-Link have not worked for me this way but I'll try to find out some cheap solution; most people probably do not have Segger J-Link.
The files from the seller are attached there as well.

thelsing:
@pavkriz Did you get anywhere further with this board or do you plan to work on this? Otherwise I would close this issue.

pavkriz:
I got SWD working with a cheap ST-Link clone and through 3 debug pins only. I'll update the instructions ASAP. My colleague has prepared a "testbed" and is going to try the library on the board.

m-byte:
[...]

PR #209, opened 2022-06-08
H8I8O

m-byte:
See #181 

This adds some new defines, a workaround for the GD32 and configurations for both H8I8O and H8C09 boards.


m-byte:
Sorry about the failing tests. There were a few changes I did after testing and they broke support for some other platforms. Those should be fixed now (everything compiles successfully on my machine)

thelsing:
Merged. Thanks.

PR #210, opened 2022-06-10
H8I8O Example

m-byte:
Firmware for the H8I8O that allows configuring pins as input, output or PWM output in a similar way to the original.


thelsing:
When you feel that this code is ready for inclusion, mark the PR as "ready for review"

PR #213, opened 2022-07-15
remove unnecessary print to console

mumpf:
This pull request just removes a print in dptconvert.cpp, which generates unexpected (and not understandable) output during debug sessions.

Regards, Waldemar


mumpf:
To be honest, I have no clue, why the tests are failing, but for sure this is not caused by my change.

Can you have a look or tell me, how I can help?

Regards,
Waldemar

PR #216, opened 2022-10-15
RP2040 EEPROM/Flash compiler flag fix

Sonnengruesser:
define compiler instructions to check if EEPROM emulation or Flash is used before checking flash size definition.


src/rp2040_arduino_platform.cpp
```diff
@@ -35,13 +35,15 @@ A RAM-buffered Flash can be use by defining USE_RP2040_LARGE_EEPROM_EMULATION
 
 #define FLASHPTR ((uint8_t*)XIP_BASE + KNX_FLASH_OFFSET)
 
+#ifndef USE_RP2040_EEPROM_EMULATION
```
SirSydom, line 38: ok

src/rp2040_arduino_platform.cpp
```diff
@@ -35,13 +35,15 @@ A RAM-buffered Flash can be use by defining USE_RP2040_LARGE_EEPROM_EMULATION
...
 #if KNX_FLASH_SIZE%4096
 #error "KNX_FLASH_SIZE must be multiple of 4096"
 #endif
 
 #if KNX_FLASH_OFFSET%4096
 #error "KNX_FLASH_OFFSET must be multiple of 4096"
 #endif
+#endif
```
SirSydom, line 46: ok

src/rp2040_arduino_platform.h
```diff
@@ -4,10 +4,15 @@
 
 #ifdef ARDUINO_ARCH_RP2040
 
+#define USE_RP2040_EEPROM_EMULATION
```
SirSydom, line 7: not ok. you cannot define USE_RP2040_EEPROM_EMULATION as default here.
In that case it is not anymore possible to undef it without changing the stacks code.
the defines used here are meant to be defined as compile flags in your build environment.

src/rp2040_arduino_platform.h
```diff
@@ -4,10 +4,15 @@
...
+#warning "using RP2040 EEPROM emulation"
+
+#ifndef USE_RP2040_EEPROM_EMULATION
```
SirSydom, line 10: ok

src/rp2040_arduino_platform.h
```diff
@@ -4,10 +4,15 @@
...
 #ifndef KNX_FLASH_OFFSET
 #define KNX_FLASH_OFFSET 0x180000   // 1.5MiB
 #pragma warning "KNX_FLASH_OFFSET not defined, using 0x180000"
 #endif
+#endif
```
SirSydom, line 15: ok

SirSydom (changes requested):

Sonnengruesser (commented):
review comments by @SirSydom checked and define removed.

src/rp2040_arduino_platform.h
Sonnengruesser: agreed and understood.
I'm using the Arduino IDE without the possibility to define compile flags. Where would you put this define then?
In general: I'm transitioning a sketch from SAMD21 to RP2040 and have to adapt compile flags in several .h-files. That's not really user friendly and requires deep knowledge of the stack.

SirSydom: > and have to adapt compile flags in several .h-files. That's not really user friendly
yes, and it is no good style at all.
That is a big flaw of the Arduino IDE and thats one reason I use Visual Studio Code with the Arduino or with Plattform I/O Extension. In both it is possible to use Compiler flags.
For OpenKNX, we choose Plattform I/O as part of our unified tool chain (altough the P2040 core is not yet officially supported). There it looks like that:

```
[env]
framework = arduino
build_flags = 
  -include "src\SEN-UP1-8xTH.h"
  -include "src\SEN-UP1-8xTHHardware.h"
  -D THPSENSORMODULE
  ; optimize knx stack
  -D SMALL_GROUPOBJECT
  ; -D USE_BINSEARCH
  ; -D KNX_AUTO_ADAPT
  ; -D KNX_WAIT_FOR_ADDR
  -D MASK_VERSION=0x07B0
  ; use flash directly
  ;flash start at 1.5MiB
  -D KNX_FLASH_SIZE=0x4000
  -D KNX_FLASH_OFFSET=0x180000 
  -D USERDATA_SAVE_SIZE=1004
  ; use EEPROM emulation
  ;-D DEBUG_TIMING
  ;-D LOGIC_TRACE
  -D WAITFORSERIAL=1
  -Wno-unknown-pragmas 
  -Wno-switch
  ; debug timing in knx stack
```


Maybe in Arduino IDE 2.0 it is also possible to use compiler flags - but that is beyond my knowledge.

SirSydom (approved):

PR #217, opened 2022-10-15
Samd51 flash storage compatibility

OnlineCaveman:
I wanted to use the repository on a SAMD51 and noticed the flash memory routines were only for SAMD21 so here are my proposed changes for SAMD51 compatibility.

Since I have no SAMD21 board available at the moment, I can only verify the code on a SAMD51 platform.


SirSydom (commented):
maybe in that case a samd51_plattform - derived from the samd_plattform would be a better solution then extensive ifdef usage...

OnlineCaveman:
Good suggestion. For now my time is limited. I will change this to a draft and come back to it.

src/knx/knx_value.cpp
```diff
@@ -1,562 +1,562 @@
...
+{
+    _value.doubleValue = value;
+    _type = DoubleType;
+    return *this;
+}
+
+KNXValue::operator float() const
+{
+    return doubleValue();
 }
```
thelsing, line 562: Did you change anything here? Where the line endings wrong?

src/samd_platform.cpp
```diff
@@ -1,6 +1,10 @@
+#if(defined(__SAMD21E17A__) || \
+    defined(__SAMD21G18A__) || \
+    defined(__SAMD21E18A__) || \
+    defined(__SAMD21J18A__))
+
 #include "samd_platform.h"
 
-#ifdef ARDUINO_ARCH_SAMD
```
thelsing, line 3: Any reason to not also put
 #if (defined(__SAMD21E17A__) || \
     defined(__SAMD21G18A__) || \
     defined(__SAMD21E18A__) || \
     defined(__SAMD21J18A__)) 
here?

src/knx/knx_value.cpp
OnlineCaveman: No, nothing changed as far as I know. Was surprised to see it being included in the push.

src/samd_platform.cpp
OnlineCaveman: No, copy paste ended up at the start of the file. I will move them after the include samd_platform.h to be consistent with the other files.

OnlineCaveman:
Base work for W5500 support made. Errors in macro expansion getbyte

Intellisense message:
#define htonl(x) ( (getbyte(x,0)<<24) | (getbyte(x,1)<<16) | (getbyte(x,2)<<8) | getbyte(x,3) )
Expands to:
( (getbyte(io->_platform.currentIpAddress(),0)<<24) | (getbyte(io->_platform.currentIpAddress(),1)<<16) | (getbyte(io->_platform.currentIpAddress(),2)<<8) | getbyte(io->_platform.currentIpAddress(),3) )
expression must be an lvalue or a function designatorC/C++(158)

Build error:
In file included from src\knx\ip_parameter_object.cpp:5:
src\knx\ip_parameter_object.cpp: In lambda function:
src\knx\bits.h:9:39: error: lvalue required as unary '&' operand
    9 | #define getbyte(x,n) (*(((uint8_t*)&(x))+n))
      |                                       ^
src\knx\bits.h:11:21: note: in expansion of macro 'getbyte'
   11 | #define htonl(x) ( (getbyte(x,0)<<24) | (getbyte(x,1)<<16) | (getbyte(x,2)<<8) | getbyte(x,3) )
      |                     ^~~~~~~
src\knx\ip_parameter_object.cpp:49:25: note: in expansion of macro 'htonl'
   49 |                 pushInt(htonl(io->_platform.currentIpAddress()), data);
      |                         ^~~~~

Anyone a clue as why this is the case? The samd51platform::currentIpAddress() returns Ethernet.localIP() analogue to WiFi.localIP() as used in the esp_platform and esp32_platform cases, both of which compile fine.

thelsing:
Maybe we should put ARCH defines around htonl and friends so that we only provide them if the platform doesn't have them already.

thelsing (changes requested):
Please get rid of the changes of knx_value or at least investigate what  changed. (If this was the only file with different line endings this change would be ok)

In knx.demo is too much stuff now. The setup code for Ethernet should be moved to the Platform. The esp-Platforms have examples on how to do this.

Also please don't remove the samd-zero env from platform.ini. You are free to add the new env  for adafruit_feather_m4.

thelsing:
I just merged this manually into devel. Thanks for your contribution.

PR #218, opened 2022-10-17
Feature175 binarysearch

mumpf:
Implements a binary search algorithm for GroupAdresses and GroupObjects.
Supports especially sending ACK in time during groupaddress base communication.

You have to #define BIN_SEARCH during compilation to use this functionality. Is just important if your firmware needs to supprt many GA (like 1000 or so).

Coding and corner cases (finding first, last, middle and no element) were tested. 

@SirSydom: I took your coding and changed/modified it slightliy. Thank you.

PR #220, opened 2022-10-26
reimplemented binary search for translateAsap()

mumpf:
- new prepareBinarySearch() called as soon as association_table changes
- fallback to linear search if ETS does not provide sorted ASAP entries
- activated with #define  BIN_SEARCH

This closes existing functionality gaps found after closing of #218. It was carefully testes on 4 different devices and seems to work as expected. I debugged the coding also taking some corner cases into account.

Please review and merge these changes.

Regards,
Waldemar

### DIFF
--- a/src/knx/association_table_object.cpp
+++ b/src/knx/association_table_object.cpp
@@ -39,10 +39,64 @@ uint16_t AssociationTableObject::getASAP(uint16_t idx)
     return ntohs(_tableData[2 * idx + 2]);
 }
 
+// after any table change the table is checked if it allows
+// binary search access. If not, sortedEntryCount stays 0, 
+// otherwise sortedEntryCount represents size of bin search array 
+void AssociationTableObject::prepareBinarySearch()
+{
+    sortedEntryCount = 0;
+#ifdef USE_BINSEARCH
+    uint16_t lastASAP = 0;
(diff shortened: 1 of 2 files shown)


src/knx/association_table_object.cpp
```diff
@@ -39,46 +39,94 @@ uint16_t AssociationTableObject::getASAP(uint16_t idx)
...
 int32_t AssociationTableObject::translateAsap(uint16_t asap)
 {
     uint16_t entries = entryCount();
-    #ifdef USE_BINSEARCH
-    uint16_t low,high,i;
-    low = 0;
-    high = entries-1;
-
-    while(low <= high)
+    if (sortedEntryCount)
```
thelsing, line 107: I think you should add a comment here that you rely on strong monotone sorting here, and we know it is because of the prepareBinSearch method.

thelsing (approved):
Looks fine. How about we get rid of the USE_BINSEARCH compiler flag? Is it a difference in code size?

src/knx/association_table_object.cpp
mumpf: Hi, I added the required comments. I hope they explain my approach.

Regards,
Waldemar

PR #236, opened 2023-03-24
Added FunctionProperty Callback

thewhobox:
Added a callback for FunctionProperty.
Now we can use it in OGM-Common Modules so the ETS can communicate with the device (via Buttonscripts).

Also fixed the length.
(Request was always 1 byte to short, and Response 1 byte to long)


thelsing:
PRs should always made against master branch. Can you please resubmit it against master?

PR #238, opened 2023-04-04
Added FunctionProperty Callback

thewhobox:
Added a callback for FunctionProperty.
Now we can use it in OGM-Common Modules so the ETS can communicate with the device (via Buttonscripts).

Also fixed the length.
(Request was always 1 byte to short, and Response 1 byte to long)


thewhobox:
Fixed and tested ADCResponse.

If you read now the DeviceInfo via ETS it won't stock because Stack doesn't send an answer.
It now responds an "error" (just an empty adc value, like specified in spec) to ets so it will continue.

PR #240, opened 2023-04-24
Fix IntelliSense w/ PlatformIO IDE on Linux

ginkel:
IntelliSense was broken in downstream projects in VS Code w/ the PlatformIO IDE plugin on Linux (as `inet.h` does not end up on VS Code's include path).

This PR fixes this by removing the include from `knx/bits.h` as it is also included directly by `linux_platform.cpp` and removing the include solves the IntelliSense issues.

### DIFF
--- a/src/knx/bits.h
+++ b/src/knx/bits.h
@@ -4,10 +4,9 @@
 #include <cstdint>
 
 #if defined(__linux__)
-#include <arpa/inet.h>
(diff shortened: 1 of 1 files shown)


thelsing:
You broke the native build. Please fix that.

thelsing:
Having thought about it a bit I think it's strange that you want to remove the include. PlatformIO intellisense shouldn't define __linux__ if you are on an embedded platform. It's probably easier to fix it vscode than changing the code.

Regarding to where to change: less changes are mostly better ;)

ginkel:
I figured out a workaround that does not require any code changes in `knx`, so I am closing this PR for now.

PR #241, opened 2023-05-04
fix: prevent use of uninitialized _savedAsapXXX variables

mumpf:
This fix prevents a hanging situation in the stack. 
In rare cases it might happen, that confirm mesages are send before _savedAsapXXX is set. 
One of these rare cases: [this coding is entered](https://github.com/OpenKNX/knx/blob/10266a2966cb5923d8774199bb5b3bc6c756e3fa/src/knx/tpuart_data_link_layer.cpp#L455-L464), the sendBuffer contains an incorrect telegram and this leads to a send telegram with an ASAP = 0, which does not exist. 

This correction improves robustness and has no other side effects as far as I see. I tested it for more than 2 weeks in a productive system with thousands of telegrams. 

Please accept this pull request.

### DIFF
--- a/src/knx/application_layer.cpp
+++ b/src/knx/application_layer.cpp
@@ -82,13 +82,25 @@ void ApplicationLayer::dataGroupConfirm(AckType ack, HopCountType hopType, Prior
     switch (apdu.type())
     {
     case GroupValueRead:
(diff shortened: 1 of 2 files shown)


thelsing:
Merged thanks.

PR #256, opened 2023-08-11
add support for external flash system by using callbacks

traxanos:


mgeramb:
@thelsing Could you check, if an overload of the memory handling in this way would be acceptable to take over? It would be used in my https://github.com/OpenKNX/OAM-SmartHomeBridge/tree/main project and improves the memory handling to use a partition with the native esp sdk functions instead of the EEPROM emulation.

traxanos:
i have merge our internal devel branch so die MR was also updated :(

but only this commit is relevant: https://github.com/thelsing/knx/pull/256/commits/5cdd47d8f9be3ad16069c38afcc2f58cc868bac2

thelsing:
I don't think KNX_FLASH_CALLBACK define is needed. Overhead is only the size of three pointers. What do others think?

thelsing:
Also please remove the redundant line that CodeFactor doesn't like.

mgeramb:
> I don't think KNX_FLASH_CALLBACK define is needed. Overhead is only the size of three pointers. What do others think?

There would be only one advantage to keep the define. The code requires currently a define of KNX_FLASH_SIZE. But this is not used in the case of using callback. But to fix this, more changes would necessary to exclude all KNX_FLASH_SIZE usages in the case that KNX_FLASH_CALLBACK is used. 
Another possibility would be to set the KNX_FLASH_SIZE also with a function, to be downward compatible the default value of the flash size variable could be set with KNX_FLASH_SIZE if this define was set. In my opinion, this solution would be the smartes and the additional memory footprint would be to have the callback points and the variable which stores the KNX_FLASH_SIZE

traxanos:
actually i am not a fan for using "_memoryType". Background is that the implementation does not change at runtime. i would even like to use KNX_FLASH_CALLBACK to exclude much more incl. KNX_FLASH_SIZE, which is superfluous. furthermore KNX_FLASH_CALLBACK is currently also used by OPENKNX to switch from the old to the new method.

traxanos:
so i have update my merge request.

PR #257, opened 2023-08-22
FIX: float16ToPayload

mumpf:
- all DPT9 numbers of the form (2^n)/100, n>10 (first value is 20.48) were converted to a payload of 0
- in other words: 20.48; 40.96; 81.92; etc. were send as 0!
Fix is tested and works fine. First time I got a bug report regarding this error was 3 years ago. At this time I couldn't reproduce...

Please merge this fix. Thanks. Waldemar

### DIFF
--- a/src/knx/dptconvert.cpp
+++ b/src/knx/dptconvert.cpp
@@ -1749,8 +1749,8 @@ void float16ToPayload(uint8_t* payload, size_t payload_length, int index, double
     value *= 100.0;
     unsigned short exponent = 0;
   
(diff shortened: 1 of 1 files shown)


mumpf:
Just a remark: I tested the fix also with negative numbers and also the reverse function float16FromPayload (in case there is this error, too). Everything works now fine!

mumpf:
After additional tests there were still some rounding problems. Improved correction and did a test with full numeric range of DPT9. Finally, just value 0 results in a payload 0.

PR #258, opened 2023-09-07
Allow Setting Value of Group-Object with Sending Changed Converted Value Only

cornelius-koepp:
Will check based on changed value **after** conversion.

Calling `valueModifiedSend(value, type)` with different values will only send, if value represented in given dpt is not the same.

### DIFF
--- a/src/knx/group_object.cpp
+++ b/src/knx/group_object.cpp
@@ -280,7 +280,7 @@ void GroupObject::valueNoSend(const KNXValue& value, const Dpt& type)
     KNX_Encode_Value(value, _data, _dataLength, type);
 }
 
-bool GroupObject::valueSendChangedOnly(const KNXValue& value, const Dpt& type)
+bool GroupObject::valueModifiedSend(const KNXValue& value, const Dpt& type)
 {
     // save current value
     const uint8_t oldLength = _dataLength;
--- a/src/knx/group_object.cpp
+++ b/src/knx/group_object.cpp
@@ -280,7 +280,7 @@ void GroupObject::valueNoSend(const KNXValue& value, const Dpt& type)
     KNX_Encode_Value(value, _data, _dataLength, type);
 }
 
-void GroupObject::valueSendChangedOnly(const KNXValue& value, const Dpt& type)
+bool GroupObject::valueSendChangedOnly(const KNXValue& value, const Dpt& type)
 {
     // save current value
     const uint8_t oldLength = _dataLength;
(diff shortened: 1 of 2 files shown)


thelsing:
Is there any reason that you marked this PR as draft? Do you still want to change something?

src/knx/group_object.cpp
```diff
@@ -279,3 +279,30 @@ void GroupObject::valueNoSend(const KNXValue& value, const Dpt& type)
 
     KNX_Encode_Value(value, _data, _dataLength, type);
 }
+
+bool GroupObject::valueModifiedSend(const KNXValue& value, const Dpt& type, const bool forceSend /*= false*/)
+{
+    if (_commFlag == Uninitialized)
```
mumpf, line 285: add forceSend here:
`if (forceSend || _commFlag == Uninitialized)`

src/knx/group_object.cpp
```diff
@@ -279,3 +279,30 @@ void GroupObject::valueNoSend(const KNXValue& value, const Dpt& type)
...
+        uint8_t newData[_dataLength];
+        KNX_Encode_Value(value, newData, _dataLength, type);
+
+        // check for change in converted value / update value on change only
+        const bool dataChanged = memcmp(_data, newData, _dataLength);
+        if (dataChanged)
+            memcpy(_data, newData, _dataLength);
+
+        // trigger sending
+        if (dataChanged || forceSend)
```
mumpf, line 303: Remove this if and execute objectWritten() directly after memcpy.

mumpf (commented):
In my opinion the forceSend argument should not be there, because there is already the value method, which clearly sends a value.

But if you add forceSend, it should be handled together with Uninitialized.

cornelius-koepp:
> because there is already the value method, which clearly sends a value.

The `value(..)`-method should be preferred when the knowlege of object value changed is **not** relevant.

> forceSend [...] should be handled together with Uninitialized.

By handling the same way as `Uninitialized`, there would be no feeback of object value change and no advantage in comparison to `value(..)`.

src/knx/group_object.cpp
cornelius-koepp: Would change (intended) behaviour: The return value should always indicate a changed value/state.

Starting from `Uninitialized` or setting the first value (assumption: no defined value and response to read requests) there is always a change. For following updates it depends on existing and new value.

src/knx/group_object.cpp
cornelius-koepp: Would change (intended) behaviour: The return value should always indicate a changed value/state, independent of writing to bus.

cornelius-koepp:
> Is there any reason that you marked this PR as draft? Do you still want to change something?

Yes, the 'draft' state was intended, to have the room for optimizations, feedback (for this hopefully generically usable extension of public group-object-api) and further testing. Optimization and feedback seen yesterday.

cornelius-koepp:
Replaced by #272 with checking for update only.

PR #261, opened 2023-10-25
remove byteorder macros for rp2040 platform

traxanos:
remove byteorder macros for rp2040 platform as they are already included in lwip of sdk.


Ing-Dom (approved):
no objections

PR #264, opened 2023-12-17
Add activity callback function

Ing-Dom:
Add the possibility to have a callback function signaling knx activity (for e.g. flashing LEDs)

Can be enabled with KNX_ACTIVITYCALLBACK

### DIFF
--- a/src/knx/data_link_layer.h
+++ b/src/knx/data_link_layer.h
@@ -18,7 +18,7 @@ protected:
     ActivityCallback _activityCallback = nullptr;
 public:
     virtual ~DataLinkLayerCallbacks() = default;
-    virtual void Activity(uint8_t info);
+    virtual void activity(uint8_t info);
     virtual void setActivityCallback(ActivityCallback activityCallback);
 };
 
--- a/src/knx/data_link_layer.h
+++ b/src/knx/data_link_layer.h
@@ -42,5 +42,8 @@ class DataLinkLayer
     Platform& _platform;
 #ifdef USE_CEMI_SERVER
     CemiServer* _cemiServer;
-#endif    
+#endif
+#ifdef KNX_ACTIVITYCALLBACK
+    uint8_t _netIndex = 0;
+#endif
 };
--- a/src/knx/data_link_layer.h
+++ b/src/knx/data_link_layer.h
@@ -10,6 +10,18 @@
 
 class Platform;
 
+typedef void (*ActivityCallback)(uint8_t info);
+
+class DataLinkLayerCallbacks
+{
+protected:
+    ActivityCallback _activityCallback = nullptr;
+public:
+    virtual ~DataLinkLayerCallbacks() = default;
+    virtual void Activity(uint8_t info);
+    virtual void setActivityCallback(ActivityCallback activityCallback);
+};
+
 class DataLinkLayer
 {
   public:
--- a/src/knx/ip_data_link_layer.cpp
+++ b/src/knx/ip_data_link_layer.cpp
@@ -30,6 +30,9 @@ bool IpDataLinkLayer::sendFrame(CemiFrame& frame)
     if(isSendLimitReached())
         return false;
     bool success = sendBytes(packet.data(), packet.totalLength());
+#ifdef KNX_ACTIVITYCALLBACK
+        knx.Activity((_netIndex << KNX_ACTIVITYCALLBACK_NET) | (KNX_ACTIVITYCALLBACK_DIR_SEND << KNX_ACTIVITYCALLBACK_DIR));
+#endif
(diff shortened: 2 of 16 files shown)


thelsing:
For achitectional reasons I will not merge this. The knx-Facade implements the facade pattern an simplifies usage for lib users. It is completely optional. Therefore you can't use it in any of the other parts of the stack. For the same reason its global instance must not be used in the stack. 

The correct way is to add the callback to be Bau. There can be multiple Baus (in routers for example) and each can have its own callback.

thelsing (changes requested):
See comment on PR:

Ing-Dom:
yes you are right.
I'll rework it.
do you really mean bau, because the router baus in the stack are "single"
I think you mean more then one link layer do you, like in the bau091A ? (or the other coupler baus)

Ing-Dom:
Hi Thomas,

I redesigned the feature according you hints.

thelsing:
I mean more than one logical device in one physical device. My ip router for example has one PA for the router application and a different one for additional stuff like providing time to the bus. The facade was never meant for bigger stuff like openknx with people who know how the stack is structured,  but for simple sketches for typical arduino users.

src/knx/data_link_layer.h
```diff
@@ -10,6 +10,18 @@
...
 
+typedef void (*ActivityCallback)(uint8_t info);
+
+class DataLinkLayerCallbacks
+{
+protected:
+    ActivityCallback _activityCallback = nullptr;
+public:
+    virtual ~DataLinkLayerCallbacks() = default;
+    virtual void Activity(uint8_t info);
```
thelsing, line 21: Methods should start lower case.

src/knx/ip_data_link_layer.cpp
```diff
@@ -67,6 +76,10 @@ void IpDataLinkLayer::loop()
             KnxIpSearchResponse searchResponse(_ipParameters, _deviceObject);
 
             auto hpai = searchRequest.hpai();
+#ifdef KNX_ACTIVITYCALLBACK
+    if(_dllcb)
```
thelsing, line 80: Indent looks of. Is it a spaces vs. tabs issue?

src/knx/tpuart_data_link_layer.cpp
```diff
@@ -653,6 +658,10 @@ bool TpUartDataLinkLayer::sendSingleFrameByte()
     if (_TxByteCnt >= _sendBufferLength)
     {
         _TxByteCnt = 0;
+#ifdef KNX_ACTIVITYCALLBACK
+    if(_dllcb)
```
thelsing, line 662: Same indent format as above.

src/knx/tpuart_data_link_layer.h
```diff
@@ -21,8 +21,12 @@ class TpUartDataLinkLayer : public DataLinkLayer
     using DataLinkLayer::_platform;
 
   public:
+    // TpUartDataLinkLayer(DeviceObject& devObj, NetworkLayerEntity& netLayerEntity,
```
thelsing, line 24: Is there any reason to keep this?

thelsing (changes requested):
This version is fine beside some minor style issues.  Sorry for being nitpicky but I have to work with code at work were people weren't for a few years...

src/knx/tpuart_data_link_layer.cpp
Ing-Dom: fixed

src/knx/ip_data_link_layer.cpp
Ing-Dom: fixed

src/knx/data_link_layer.h
Ing-Dom: changed Activity to activity

Ing-Dom:
> I mean more than one logical device in one physical device. My ip router for example has one PA for the router application and a different one for additional stuff like providing time to the bus. The facade was never meant for bigger stuff like openknx with people who know how the stack is structured, but for simple sketches for typical arduino users.

I work on an ip router (091A) based on this stack for alomost one year (PR will come if its a little more field tested).
For this I thought about how to solve the second application (virtual knx device) but I think that will not be so easy - some sort of bus-access splitter (or however you call it) would be neccessary, right?

thelsing (approved):
All ok now.

thelsing:
> I work on an ip router (091A) based on this stack for alomost one year (PR will come if its a little more field tested). For this I thought about how to solve the second application (virtual knx device) but I think that will not be so easy - some sort of bus-access splitter (or however you call it) would be neccessary, right?

For knx-ip you should be able open more than one multicast connection from one host and use them in parallel.

PR #266, opened 2023-12-20
Add KNX-IP support for the rp2040 plattform

Ing-Dom:
this adds IP (WiFi and wired Ethernet) support to the rp2040 platform.
It includes also some optimizations for KNX-IP in general.

This PR also includes the commits from PR #265 - so it can be easily merged after that one.

### DIFF
--- a/src/knx/ip_data_link_layer.cpp
+++ b/src/knx/ip_data_link_layer.cpp
@@ -70,9 +70,17 @@ void IpDataLinkLayer::loop()
             _platform.sendBytesUniCast(hpai.ipAddress(), hpai.ipPortNumber(), searchResponse.data(), searchResponse.totalLength());
             break;
         }
+        case SearchRequestExt:
+        {
+            // FIXME, implement (not needed atm)
+            break;
+        }
         default:
+#ifdef KNX_LOG_IP
             print("Unhandled service identifier: ");
             println(code, HEX);
+#endif
+        ;
     }
 }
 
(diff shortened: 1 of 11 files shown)


src/knx/ip_data_link_layer.cpp
```diff
@@ -70,9 +70,17 @@ void IpDataLinkLayer::loop()
             _platform.sendBytesUniCast(hpai.ipAddress(), hpai.ipPortNumber(), searchResponse.data(), searchResponse.totalLength());
             break;
         }
+        case SearchRequestExt:
+        {
+            // FIXME, implement (not needed atm)
+            break;
+        }
         default:
+#ifdef KNX_LOG_IP
```
thelsing, line 79: Can we not clutter the code with different LOG_THIS, LOG_THAT, LOG_SOMETHING_ELSE? If I remember correctly DEBUG should already be used in the code.  If there is more granular logic needed. There could be more levels like INFO, WARN, ERR or something.

thelsing:
Please separate PRs in the future and please resolve the CodeFactor issue.

src/knx/ip_data_link_layer.cpp
Ing-Dom: the logging in the stack is very basic and has no levels. Its pain in the ass to debug specific stuff when you first have to comment in 100 lines.
That was the obvious, non breaking solution.
We (@traxanos and me) plan to rework this, with log levels AND specific settings that allow you to log only what you need.

PR #279, opened 2024-02-12
Fix: Group-Object `_commFlag != Uninitialized` does NOT Guarantee an Actual Initialization

cornelius-koepp:
### Issue:

The differentiation between uninitialized and initialized state can NOT be represented correctly in ComFlag alone:
> it might be in state Transmitting during a ReadRequest on startup while value is still not valid

### Solution:
Add a separate uninitialized flag to overcome the limitations of ComFlag.

### Compatibility:
 * We use MSB to store uninitialized state and keep the size of GroupObject the same saving memory resources.
 * The old uninitialized handling is not changed for compatibility reasons.

### Tested Platforms:
- [X] RP2040
### No Issues Reported
- [X] SAMD
- [X] ESP32

### DIFF
--- a/src/knx/group_object.cpp
+++ b/src/knx/group_object.cpp
@@ -12,7 +12,8 @@ GroupObjectTableObject* GroupObject::_table = 0;
 GroupObject::GroupObject()
 {
     _data = 0;
-    _commFlag = Uninitialized;
+    _commFlagEx.uninitialized = true;
+    _commFlagEx.commFlag = Uninitialized;
     _dataLength = 0;
 #ifndef SMALL_GROUPOBJECT
     _updateHandler = 0;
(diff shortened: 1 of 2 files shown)


cornelius-koepp:
No abnormalities reported for SAMD and ESP32; removing draft state.

thelsing:
This seem to be a bit complicated to me. Why not just add an _isInitialized bool to the GO?

cornelius-koepp:
> Why not just add an _isInitialized bool to the GO?
This would add 1 additional byte per GO, or 2KB additional memory requirement for 2000 GOs. Which is a relevant share on e.g. SAMD21-environments and could break existing usages.

thelsing:
This change does the same. You only hide this byte in the ComFlagEx struct. (It's probably 4 bytes for the bool though.) Or do I overlook something?

Issue #275, opened 2024-01-16
CI Pipeline broken


Ing-Dom:
Some dependency in knx-usb seems broken. I have no Clue how to fix this.
Should we remove the example from ci ?

thelsing:
Works again.

PR #281, opened 2024-02-25
Fix: CI Pipeline Failed for Plattform I/O Core 6.1.11

cornelius-koepp:
Prevents CI Pipeline Fail `*** missing SConscript file 'custom_hwids.py'`in Step **Build knx-usb** of **PlatformIO CI / build** 

> custom_hwids.py was never executed in the ci pipeline, but the missing script was just a warning. With Plattform I/O Core 6.1.11 missing scripts are errors, so the pipeline fails.


cornelius-koepp:
Will solve #275

PR #282, opened 2024-04-17
Add function to group object to send a value only if it was changed.

mgeramb:
Add a function to set and send a value only if it was changed. Similar to the function valueNoSentCompare the function is named valueCompare.


mgeramb:
The build error https://github.com/thelsing/knx/actions/runs/8726457745/job/23941764455?pr=282 is not related to the change and seems to be a general issue.

Issue #291, opened 2024-08-19
Stack stuck trying to load/unload application


embed-me:
Hello,

When I try to program the device the stack gets completely stuck. I use an RP2040 with an attached external flash (128Mbit) on QSPI. 

Platformio.ini is based on the example provided in `knx/examples/knx-demo` and looks like below:
```
[platformio]
libdeps_dir = .pio/libs
src_dir = .
default_envs = rp2040

[env:rp2040]
framework = arduino
platform = https://github.com/maxgerhardt/platform-raspberrypi.git#60d6ae8
platform_packages = framework-arduinopico @ https://github.com/earlephilhower/arduino-pico/releases/download/3.9.3/rp2040-3.9.3.zip
board = rpipico
board_build.core = earlephilhower

upload_protocol = mbed

lib_deps =
  knx=https://github.com/thelsing/knx.git  ; using master because it seems to support rp2040 and OpenKNX uses it as well
  SPI

monitor_port = COM4
monitor_speed = 115200

build_flags =
  -DMASK_VERSION=0x07B0        ; Normal Device TP1 = Twisted Pair (cable)
  -DKNX_FLASH_SIZE=0x8000      ; default = 0x1000 = 4KiB; must be multiple of 4096
  ;-DKNX_FLASH_OFFSET=0xF4000  ; default = 0x180000 = 1.5MiB; must be multiple of 4096
  -DUSE_TP_RX_QUEUE            ; Improved performance on TP by using a queue instead of immediate frame processing
  -DPIO_FRAMEWORK_ARDUINO_ENABLE_RTTI  ; according to rp2040_arduino_platform.cpp: RTTI must be set to enabled in the board options
  -DDEBUG_TP_FRAMES
  -Wuninitialized
  -Wunused-variable
  -Wno-unknown-pragmas
```

main.cpp is trivial atm and contains only the bare minimum:

```
#include <Arduino.h>
#include <knx.h>

#define PROG_LED 6
#define PROG_BTN 7
#define BUZZER_PIN 8

void foo(GroupObject &go) {
    Serial.println("foo");
}

void setup_KNX() {
  Serial.println("Setup Prog Button and Prog LED");
  pinMode(PROG_LED, OUTPUT);
  pinMode(PROG_BTN, INPUT_PULLUP);

  digitalWrite(PROG_LED, LOW);

  knx.buttonPin(PROG_BTN);
  knx.ledPin(PROG_LED);
  knx.ledPinActiveOn(HIGH);

[...]

PR #292, opened 2024-08-30
check value range before accessing group objects

embed-me:
Details documented in issue: https://github.com/thelsing/knx/issues/291


thelsing:
I added the check at the devel branch. Please check it out. In the devel branch there is proper logging. You get a WARN message now. Also not that you can have up to 64K GroupObjects so the check < UINT8_MAX is wrong.

PR #300, opened 2024-11-06
Fix Sending Wrong Values on Failed Conversion to DPT

cornelius-koepp:
> **This is an Upstream-Fix from OpenKNX.** Marked as draft as long review is pending.
> Description below is a copy of https://github.com/OpenKNX/knx/pull/30

Limitation (for redesign on [devel](https://github.com/thelsing/knx/tree/devel)): This fix will *not work* with current implementation. See separate PR #301 

---
# Issue
Setting a value which cannot be converted to the given DPT did result in unexpected behaviour.
* `valueNoSendCompare`/`valueCompare` resulted in value 0 (and returned change based on this value)
* `valueNoSend` updated state from unitialized to OK, without updating the value
* `value` wrote to GA without setting the KO value

# Reproduce
*For all cases* use a value/type combination which results in `KNX_Encode_Value` return `false`.

Example: value `500` with type `DPT_Value_1_Ucount`, or any other 

Note: DPT1 will **not** fail; ignoring all other bits.

* `valueNoSendCompare`/`valueCompare`: did always set value to 0 (no special preconditions needed beside conversion fail)
* `valueNoSend`: use before any other change of KO. No change in value expected, but read requests will answered (with undefined value)
* `value` use before any other change of KO. Will result in write telegram (with other value included) 

# Fix
Suppress updating, status-update and sending for invalid values.

# To Discuss
The fix does not change the public interface for changing/writing KO values, so there will be no feedback if conversion fails. Current methods return type `void` could be changed to `bool` to indicate success, without breaking.


thelsing:
Changing the void methods to bool to signal success seems like a good idea for me. A stable API is not that important. Constant improvement of the code is :-)

cornelius-koepp:
> Changing the void methods to bool to signal success seems like a good idea for me.
I'll prepare a draft.

See #301 for modified PR for devel.

thelsing:
Do you still want to change something or can I merge this? (PR is still draft)

cornelius-koepp:
> Do you still want to change something or can I merge this?

No need of changes anymore. 

Note: Testing was *only* conducted in Upstream/OpenKNX-Environment.

thelsing:
Merged. Thanks for your contribution.

PR #310, opened 2025-05-02
Fix crash for DPTs with less than 1byte in size

Phil1pp:
ESP8266 crashed when receiving write requests on group objects of type DPT2 or DPT3 with less than 1 byte in size.


thelsing:
Merged. Thanks.

PR #311, opened 2025-05-02
handle DPT2 and DPT3 as uint8_t in code

Phil1pp:
Return DPT2 and DPT3 as uint8_t and let the application handle the different bits.


thelsing:
Merged. Thanks.

PR #313, opened 2025-05-02
Fix ESP compiler warning

Phil1pp:
ICACHE_RAM_ATTR is deprecated in the latest ESP8266 core versions and will lead to a compilation warning. IRAM_ATTR should be used instead.


thelsing:
Merged. Thanks.

PR #315, opened 2025-05-07
Don't assign built-in LED as ProgLED by default

Phil1pp:
Allow disabling ProgLED by setting _ledPin=-1

For use cases where all GPIOs are in use or LED_BUILTIN pin is used for other functionality.

### DIFF
--- a/src/knx_facade.h
+++ b/src/knx_facade.h
@@ -28,9 +28,6 @@
         void buttonUp();
     #endif
 #elif defined(ARDUINO_ARCH_ESP32)
-    #if !defined(LED_BUILTIN)
-        #define LED_BUILTIN 13
-    #endif
     #include "esp32_platform.h"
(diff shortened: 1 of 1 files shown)


src/knx_facade.h
```diff
@@ -267,26 +258,30 @@ template <class P, class B> class KnxFacade : private SaveRestore
             _bau.deviceObject().version(value);
         }
 
-        void start()
-        {
-            if (_progLedOffCallback == 0 || _progLedOnCallback == 0)
-                pinMode(ledPin(), OUTPUT);
+	    void start()
```
thelsing, line 261: It seems that some whitespaces changed here. Please change the tabs to spaces.

Phil1pp: Sorry about that. I don't know how that happened, as my editors are usally set to use spaces instead of tabs.

thelsing:
Merged. Thank you for your contribution.

Issue #144, opened 2021-06-14
Unload application is not permanent


croghostrider:
Hi @thelsing  

your stack is amazing.

I found a bug if I unload the application over the ETS.
The address table, association table, and the application will be unloaded only until the restart of the board.

After the restart, it shows everything loaded and the device is sending telegrams.

I am not sure how to fix this.

thelsing:
I would add a method the the Memory class that writes some bytes to _data and calls _platform.commitToEeprom(); in such a way that readMemory failse [here. ](https://github.com/thelsing/knx/blob/6254fc9b6789adf13058c3f243f3becde8a69b3c/src/knx/memory.cpp#L39) Just writing 0x00 should be enough for that.

SirSydom:
unload should only unload the application but not the PA, right?

croghostrider:
it should unload everything.
See here the support article:
https://support.knx.org/hc/en-us/articles/4402998506386-Unload-Device

SirSydom:
the article differs between unloading application and unloading application and address,
so both should be possible.

thelsing:
Yes. See https://github.com/thelsing/knx/blob/22af8267c3d059fb9e63a5d8ab351e74605b2540/src/knx/bau_systemB.cpp#L150 and 3.7.1.2 in 3/5/2 of the spec.

Phil1pp:
It seems that unloading is only permanent if you do a writeMemory() before the next restart.

After loading an application, ETS will initiate a restart of the device which will call
BauSystemB::restartRequestIndication, where a writeMemory() will happen.

However for unloading an application the behaviour is different.
ETS will not initiate a restart of the device and therefore writeMemory() won't be executed.
The Unload - all procedure in Kaenx-Creator is missing the <LdCtrlRestart />
Is this by design?
Should the writeMemory() and restart be handled by the application?

thelsing:
It think calling writeMemory() on unload would be ok. It would be best if we only call it only once event though unload adress table, association table etc.  

Flash memory could probably tolerate an additional 4 write on every reconfiguration from ETS, but it would be much better if we only call writeMemory() wenn all table objects are unloaded or something similar.

Phil1pp:
I can confirm that calling `writeMemory()` in line 257 does indeed work.
By adding additional debug output I can also confirm that it is called 4 times.

However due to my limited knowledge of the stack, I don't know how to implement that it would only be called once.

thewhobox:
You simply cant do that.
[...]

PR #316, opened 2025-05-12
Permanently unload device

thewhobox:
First attempt to fix permanently unloading a device according to thelsing/knx #144

### DIFF
--- a/examples/knx-demo/.gitignore
+++ b/examples/knx-demo/.gitignore
@@ -1,0 +1,5 @@
+.pio
+.vscode/.browse.c_cpp.db*
+.vscode/c_cpp_properties.json
+.vscode/launch.json
+.vscode/ipch
--- a/src/knx/bau_systemB.cpp
+++ b/src/knx/bau_systemB.cpp
@@ -165,8 +165,8 @@ void BauSystemB::memoryRoutingTableWriteIndication(Priority priority, HopCountTy
     if (_deviceObj.verifyMode())
         memoryRoutingTableReadIndication(priority, hopType, asap, secCtrl, number, memoryAddress, data);
 
-    if(_unloadStartet != 0)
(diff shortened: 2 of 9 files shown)


thewhobox:
It's not yet tested.

It just marks tables as "wasUnloaded".
The loop in bau_systemB_device (and coupler) checks this and saves the time.
If the timeout reaches 5s it will call writeMemory().
If the device receives a memoryWrite/propertyWrite/routingTableWrite/... it will reset the timeout so it may not effect the parametration of the device.

@Phil1pp could you test it?

Phil1pp:
The timeout and the write is working, however it keeps triggering the write every 5000ms.
Some kind of writeDone or resetting of the WasUnloaded bits is missing.

thewhobox:
I resetted the flag WasUnloaded.
Should now only trigger once.

Phil1pp:
It looks good now.
I did several tests for unloading with and without physical address.
The unload was persistant after a device restart.
Also loading the application again was without a problem.

thewhobox:
Thanks for your testing.

@thelsing what do you think about the solution?

thelsing:
My first idea would be to add the timeout to the Memory class instead. So on "ClearMemory" one would set a timeout and do a writeMemory after it's reached. This seems to be a bit more centralized if we clear memory for some other reason. 

Does this seam reasonable?

Phil1pp:
After you last change it doesn't work anymore because _saveTimeout is never set.
[clearMemory()](https://github.com/thelsing/knx/pull/314) was just recently added by me and isn't involved in the unloading process.

You probably still have to check if TableObject state changed to LS_UNLOADED.

thelsing:
It's a bit less obvious how this works than I'd like it to be, but it's good enough to merge.

PR #317, opened 2025-05-14
Added support for LibreTiny IoT chips

Phil1pp:
[github.com/libretiny-eu/libretiny](https://github.com/libretiny-eu/libretiny)

Tested with **CBU** chip over WIFI
```
[env:BK7231N_ip]
platform = libretiny
board = cbu
framework = arduino
build_flags =
  -DMASK_VERSION=0x57B0
  -DSMALL_GROUPOBJECT
  -DKNX_NO_SPI
  -DKNX_FLASH_OFFSET=0x1DB000
```


Phil1pp:
I don't understand why linking fails even though **randomSeed** is defined here:
[wiring_math.cpp](https://github.com/libretiny-eu/libretiny/blob/69e7e2debee54b9e985a647ee9f4f5c2cce7411a/cores/common/arduino/src/wiring/wiring_math.cpp#L21C1-L21C35)

thewhobox:
Mabye millis() is unsigned long but the function needs uint32_t.
Because there ist no function with parameter seed as unsigned long it cant link it.

If i make these changes in line 61 it works:
``` 
uint32_t seed = millis();
randomSeed(seed);
```

Phil1pp:
Are you sure that this change should work?
I wasn't able to compile it locally and also online it failed.

I only got it to compile by adding this:
```
#if defined(LIBRETINY)
#include "wiring_math.cpp"
#endif
```

thewhobox:
Strange. I tested it locally.

Phil1pp:
I replaced `randomSeed` with `srandom` and now it works.

thelsing:
Merged. Thank you for the contribution.

PR #319, opened 2025-07-03
Add native support for ESP32 ESP IDF with example.

dzungpv:
Add native support for ESP IDF with example. Support all ESP32 family chip support by the IDF.


Noschvie:
@dzungpv Would it be possible to easily integrate this stack into the Tasmota ecosystem? What do you think?
Thanks!

PR #322, opened 2025-08-23
Fix for PR #316: Special Case Handling for Unload-Timeout

cornelius-koepp:
Unload-Timeout was introduced with #316 

`_savedTimeout==0` is defined as no timeout running. Prevent setting to `0` in the rare case `millis()==0`.


cornelius-koepp:
Broken action is addressed/solved by #325

PR #327, opened 2025-11-04
Fix: Do Not warn on Construction of DPT 14.000, 15.000, 16.000

cornelius-koepp:
DPT 14.000, DPT 15.000, DPT 16.000 are defined with Subgroup 0 in knx-spec (03_07_02 Datapoint Types v02.02.01 AS)


thelsing:
Thanks for the contribution.

